PR #21, opened 2021-08-07
Added fix bond compatible benzene, pentane, and ethanol mol2 files. Added flexible benzene FF (Trappe like)

bc118:
Added fix bond compatible benzene, pentane, and ethanol mol2 files. Added flexible benzene FF (Trappe like).
- Benzene : TraPPE like FF (Yiannourakou et. al)
- Pentane : foyer TraPPE-UA
- Ethanol : foyer OPLS-AA 

The  flexible benzene FF (TraPPE like) is really the best/only option.  The only minor issues for this is with MC:
- Requires the MC engines to set the bond lengths fixed, which were not fixed in the TraPPE like FF parameterization.
- MCCCS-MN may have a difficult time regrowing rings without fixed dihedrals and need to run longer simulations
- All MC engines need to be careful not to regrow rings thru other rings, and GOMC and Cassandra may just want/need to use a crankshaft move instead of regrow.  Note: MCCCS-MN does not have a crankshaft move.

Using the traditional TraPPE FF (Siepmann) FF for benzene requires rigid/fixed bonds, angles and dihedrals, which is likely not possible without a lot of assumption or fudging of the FF.  For example to make the TraPPE FF (Siepmann) FF work in MD for benzene we need:
- To add k-constants for all the bonds, angles and dihedrals.
- LAMMPS may be able to use fix rigid command to fix the bonds, angles, and dihedrals (https://docs.lammps.org/fix_rigid.html).  However, LAMMPS can not shake linear molecules C-C atoms/beads. See the LAMMPS statement in the shake function (https://docs.lammps.org/fix_shake.html).  Specifically, The LAMMPS documentation says "This means water molecules or CH2 or CH3 groups may be constrained, but not all the C-C backbone bonds of a long polymer chain".  Therefore, it appears that shaking/fixing the bonds and angles in all Trappe-UA FFs is not possible, at least for our listed molecules. 
- Adding different and very large dihedral k-constants to pseudo fix them in MD.

**Basicly, MD will not be fixing bonds (except water and hydrogens if 2ft/step), while MC will fix all bonds regardess of the FF parameterization.**  This seems to be the least fudging/FF manipulating way to do this.

Can someone please double check the benzene FF with the paper and Excel sheet below, which I used to get, calc, and enter the data?

**Attached FF paper and my Excel calcs to put in foyer XML:**

[United_atom_forcefield_for_vapor_liquid_equilibrium_VLE_properties_of_cyclic_and_polycyclic_compounds_from_Monte_Carlo_simulations.pdf](https://github.com/mosdef-hub/reproducibility_study/files/6950001/United_atom_forcefield_for_vapor_liquid_equilibrium_VLE_properties_of_cyclic_and_polycyclic_compounds_from_Monte_Carlo_simulations.pdf)

[FF_conversion.xlsx](https://github.com/mosdef-hub/reproducibility_study/files/6949998/FF_conversion.xlsx)

Note: Methane does not work to import as a mol2 into mbuild, as there is an issue with the mol2 readers for single atom molecules, or a least building in VMD.  GOMC need to input Methane as follows to get it to work, with the residue name on the top and the "_CH4" as a child (see code below).

**Methane or single atom molecule method required for GOMC **
```python
Molecule_A = mb.Compound(name="MET")

Molecule_A_molecule = mb.Compound(name="_CH4")

Molecule_A.add(Molecule_A_molecule, inherit_periodicity=False)
```


justinGilmer:
@bc118 can you check #47 and see if you minimize a system with constraints on the `PentaneUA` object, does it have the correct dimensions and such. I think since the pentaneUA is a natural extension of the polymer building class, that we might be ok leveraging that. @jennyfothergill  and others, is there a way yall prefer to do this? For benzeneUA i think loading in from a MOL2 might be best, but im curious on others opinions.

justinGilmer (commented):
Once we have a bit more of a discussion on mol2 vs mbuild routines for the construction, then I think this is good to go.

bc118:
I believe I check most  if not all, of these mol2 molecules when I tested energy minimize.  So I am confident they are ok.

bc118:
I believe we should just use the mol2 files as they are already done with the exact bond lengths to 4-5 decimal places, which were set my running GOMC regrow moves (except benzene calc only as easy to do in a plane).  Using these mo2 files as is will allow us to get the sooner.

justinGilmer (approved):

Issue #4, opened 2021-07-21
Analysis TO-DOs


justinGilmer:
- [ ] Fully define energy breakdowns for each simulation engine
- [x] Pick/write RDF analysis code
- [x] Pick/write diffusion coefficient code

jennyfothergill:
As for diffusion coefficient, we could use [freud's MSD module](https://freud.readthedocs.io/en/latest/modules/msd.html). Looks like the diffusion coefficient can be estimated as 1/6 the slope of the MSD vs t. ([source](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3193570/))

justinGilmer:
Oops, I missed this comment, sorry!

I more so meant that we need energy term decomposition in case we need to check energies between engines. Like breaking up potential energy into: NB (coul, LJ, etc), Bonded, etc.

Although that was more for the case if we had any issues with our properties. But I think at the moment this might not be necessary.

PR #23, opened 2021-08-09
Add RDF analysis

jennyfothergill:
I'm opening this mainly to start discussion
- [x] How should we handle the system creation for each engine's output?

- [x] I'm not so familiar with signac dashboard--what is the best way to display/save the result?

begins to address issue #4

### DIFF
--- a/project/src/analysis/rdf.py
+++ b/project/src/analysis/rdf.py
@@ -1,0 +1,36 @@
+"""Facilitate the calculation of the RDF from simulation data."""
+
+import freud
+import matplotlib.pyplot as plt
+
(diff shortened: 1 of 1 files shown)


justinGilmer:
Looking great so far. I think we might want to also have a project.py file for our analysis set as well. Then we can load in these analysis modules like RDF there, and then output the rdf files to disk in each state point folder in the workspace.

jennyfothergill:
yeah, I agree! and it's weird because I actually added that 🤔 ... I'm confused why this PR isn't updating with my branch

jennyfothergill:
see #35

PR #29, opened 2021-08-09
Feat/update env files

justinGilmer:
Previously, the packages listed in `environment.yml` were included with
no thought about order.

This can make trying to find a package without a regex engine quite
difficult. The environment.yml file is now alphabetized.

An additional engines.yml has been added to include the
conda-installable engines. If the user chooses to install these, they
can update their environment by:

```bash
conda activate mosdef-study38
mamba env update -f engines.yml
```

NOTE: lammps is currently an older version that I do not recommend using
for production data, use to test input scripts.


jennyfothergill (approved):
LGTM! I was able to create and update the environment as you said. (I'm on OSX, might be good to have at least one linux user double check.)

chrisjonesBSU (approved):
Creating and updating the environment works for me on Linux

PR #31, opened 2021-08-10
Add instructions for environment setup and signac project

### LINKED ISSUE #24: Add user instructions for how to run init/project
The signac structure is awesome, but a little higher level than I have used before--let's add instructions for folks who may not be familiar with how to initialize/run this project with signac.

jennyfothergill:
fixes #24

### DIFF
--- a/project/README.md
+++ b/project/README.md
@@ -1,12 +1,31 @@
+## You are here
 Main location for reproducibility study.
 
-Layout is as below (subject to change):
+## How to use
+All commands are intended to be run from this location with the `mosdef-study38` environment active.
+To initialize the project workspace:
+```bash
+python init.py
+```
+You will see the empty workspace folder populated with the hashed job directories.
+
+Each engine directory in `src/engines` contains its own `project.py` file. To run the simulations for a specific engine (in the following example we use hoomd), you would run:
+```bash
+python src/engines/hoomd/project.py run
+```
+
+Finally, to run the analysis of the simulation output you would run:
+```bash
+python project-analysis.py run
+```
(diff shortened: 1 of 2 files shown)


project/README.md
```diff
@@ -1,12 +1,31 @@
+## You are here
 Main location for reproducibility study.
 
-Layout is as below (subject to change):
+## How to use
+All commands are intended to be run from this location with the `mosdef-study38` environment active.
```
justinGilmer, line 5: Let’s add the bash command to activate the conda env here. Since it is the only time we will need to list it.

project/README.md
```diff
@@ -1,12 +1,31 @@
...
+To initialize the project workspace:
+```bash
+python init.py
+```
+You will see the empty workspace folder populated with the hashed job directories.
+
+Each engine directory in `src/engines` contains its own `project.py` file. To run the simulations for a specific engine (in the following example we use hoomd), you would run:
+```bash
```
justinGilmer, line 13: I think we should first show using the `status` command first.

justinGilmer (changes requested):
Two small changes and then this LGTM! Thanks, @jennyfothergill !

justinGilmer (approved):

project/README.md
jennyfothergill: good point! I guess we don't know if users will use submit or run anyway...
I'll also link to the signac cli docs

project/README.md
jennyfothergill: you got it! I actually added it in my most recent commits when I updated with instructions for #29

Issue #28, opened 2021-08-09
How to collate/combine output data


jennyfothergill:
Each engine-specific project.py file will create simulation/analysis output, how should we combine all these outputs?
Some discussion about using zenodo: Would we just upload each job directory (as the hashed name will be the same across systems) as a separate entry to zenodo?

justinGilmer:
Zenodo does let you mark and upload datasets in a different way compared to a github repo. Although each dataset is limited to 50GB. But we can upload as many datasets as we need to support this. See [the help/FAQ page for additional info](https://help.zenodo.org/)

Each dataset will get its own zenodo doi. I will play around and see if I can make a few archived versions of our workspace later and try to merge them. I do not forsee this becoming an issue though. Especially since each dataset will be focused on different workspace statepoints anyway.

jennyfothergill:
OK, very cool! I see that on the zenodo website it is easy to upload a file, but I am not able to upload a folder. [This is a known issue](https://github.com/zenodo/zenodo/issues/1089).  It seems like the solution might be to zip the workspace folder. 
We should make sure to have a bash script or something similar to handle unzipping and syncing the workspaces.

tcmoore3:
This is something we should push back until we're actually ready to upload our dataset.

Issue #25, opened 2021-08-09
Sanitize simulation outputs


jennyfothergill:
From chat with @tcmoore3 and @justinGilmer:
Each engine-specific project.py file should have an additional operation--one which would sanitize the the log (containing energy and other property information) and trajectory (containing the positions, (and types?, bonds?) box information) into a standard format (perhaps a txt or npy file) 
This way, the analysis project can operate on the entire workspace and look for standard output files.

We should discuss what all information should be saved and what format to use.

jennyfothergill:
use `np.genfromtxt(names=True)` so that order doesn't matter
use space as delimiter
filename = "log.txt"
for MD we should save:
timestep
potential energy
kinetic energy
pressure
temperature
density or box length?

jennyfothergill:
I am closing this issue as it is documented in the project board

PR #35, opened 2021-08-10
Add RDF analysis

jennyfothergill:
begins to address #4 

(not sure why, but perhaps the shift from public to private borked my previous PR #23) 

- [x] add units to plot and docstring
- [x] add unit test


project/src/analysis/rdf.py
```diff
@@ -0,0 +1,35 @@
...
+import gsd.hoomd
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+def gsd_rdf(job):
+    """Compute the RDF given a GSD file."""
+    gsdfile = job.fn("trajectory.gsd")
+
+    # TODO check that these are reasonable values
```
justinGilmer, line 14: Might want to make these function arguments with these as the default values?

justinGilmer (approved):
Small comment, but looks good overall!

project/src/analysis/rdf.py
jennyfothergill: just to make sure I understand--do you mean something like
```python
def gsd_rdf(job, fname="trajectory.gsd"):
    """Compute the RDF given a GSD file."""
    gsdfile = job.fn(fname)
```
?

project/src/analysis/rdf.py
```diff
@@ -0,0 +1,35 @@
...
+    frames = 10
+    bins = 50
+    r_min = 0.5
+    r_max = 2.5
```
ramanishsingh, line 18: Shall we change r_max to the (minimum (list of side length in each prod traj frame))/2-0.01 ?

project/src/analysis/rdf.py
jennyfothergill: On further thought, I think you mean
```
def gsd_rdf(job, frames = 10, bins = 50, r_min = 0.5, r_max = 2.5):
```
and I agree!

justinGilmer: Yeah this is what I meant!

project/src/analysis/rdf.py
jennyfothergill: As the longest possible r_max, that is a good place to start! Depending on how big our systems get, we may want to go smaller, but I agree that we should have this as the default.

ramanishsingh: I think freud complains when r_max > L/2. But we can confirm that once we start the analysis.

jennyfothergill: Yes, you are correct. I have updated this code to use a default r_max based on box lengths
(minimum box length * 0.45)

ramanishsingh (approved):

chrisjonesBSU:
> LGTM, but one thought. Should the `run_analysis` operation have some requirement that the simulation is finished rather than just checking that there is a `trajectory.gsd` file?

I guess I'm not considering the step that sanitizes the outputs (#25 #28), which would occur after a simulation finishes and before `run_analysis`.  So, maybe checking that a simulation is finished isn't required here.

jennyfothergill:
> > LGTM, but one thought. Should the `run_analysis` operation have some requirement that the simulation is finished rather than just checking that there is a `trajectory.gsd` file?
> 
> I guess I'm not considering the step that sanitizes the outputs (#25 #28), which would occur after a simulation finishes and before `run_analysis`. So, maybe checking that a simulation is finished isn't required here.

yeah, exactly. Because each output should be sanitized before analysis, the trajectory file existing is a proxy for the simulation being finished. 
Once we add equilibration/decorrelation to our workflows that may be added as a precondition.

jennyfothergill:
This should be ready for review. I'd like feedback on how to improve the unit test.

justinGilmer:
Overall this looks good, but maybe we can add in a RDF of a system we can verify through another calculation? Like make a similar file output from a xtal system and a liq system? We can change the methods up a bit like this to make it easier

```python3
def gsd_rdf(job, frames=10, stride=1, bins=50, r_min=0.5, r_max=None):
    return _gsd_rdf_(job.trj, ...)

def _gsd_rdf_(trj, frames, stride, bins, r_min, r_max):
    # calculations here
```

justinGilmer:
Then we can test the output of the method without relying on a job at first.

justinGilmer (changes requested):
Overall LGTM, i added one small suggestion to maybe add a known rdf and gsd file to then compare the output. Sort of hacky, but i cant really think of a way to test the rdf more.

project/tests/base_test.py
```diff
@@ -13,7 +17,50 @@ def spceff(self, name="spce.xml"):
...
+    def tmp_project(self):
+        with signac.TemporaryProject(name="test") as p:
+            return p
+
+    @pytest.fixture
+    def tmp_job(self, tmp_project, statepoint={"a": 0}):
+        tmp_project.open_job(statepoint).init()
+        for job in tmp_project:
+            return job
+
```
justinGilmer, line 35: This needs the `@pytest.fixture` decorator

justinGilmer (changes requested):

project/tests/base_test.py
justinGilmer: This is for the `trappe_ua` fixture. It did not select multiple lines. Whoops.

jennyfothergill: yeah, that's my bad! I've fixed it in a later commit.

jennyfothergill:
@justinGilmer thank you for your suggestions! I've taken a shot at implementing them.

justinGilmer (approved):
This is excellent @jennyfothergill ! (I was unaware of the `TemporaryProject` ability too! 

This LGTM!

PR #36, opened 2021-08-10
complete project.py for mcccs

ramanishsingh:
Complete project.py for methane
Added slurm* to .gitignore
Added molecular weights to init.py


project/src/engine_input/mcccs/pentane/fort.4.melt~HEAD
```diff
@@ -0,0 +1,258 @@
+
```
justinGilmer, line 1: Is this file expected to be named: `fort.4.melt~HEAD` Looks like it might have been created by accident?

justinGilmer (changes requested):
One small question to answer and then we can merge this. Thanks!

project/src/engine_input/mcccs/pentane/fort.4.melt~HEAD
ramanishsingh: Thanks. Yes, something weird happened. The file is not required and is deleted.

justinGilmer (approved):

PR #37, opened 2021-08-10
Create github action for unit testing

justinGilmer:


.github/workflows/main.yml
```diff
@@ -0,0 +1,44 @@
+name: CI
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
jennyfothergill, line 9:
```suggestion
  push:
    branches: [ main ]
    paths:                                                                      
      # run the tests when any python file is changed                           
      - '**.py'   
  pull_request:
    branches: [ main ]
    paths:                                                                      
      # run the tests when any python file is changed                           
      - '**.py'   
```
Adding this will prevent pytest from running for changes to a README, for example. (There might be more files/types we want to include though.)

.github/workflows/main.yml
```diff
@@ -0,0 +1,44 @@
...
+        python-version: ["3.8"]
+        os: ["ubuntu", "macos"]
+        include:
+          - os: ubuntu
+            environment-file: environment.yml
+            miniforge-variant: Mambaforge
+            miniforge-version: 4.9.2-4
+          - os: macos
+            environment-file: environment.yml
+            miniforge-variant: Mambaforge-pypy3
```
jennyfothergill, line 31: this is cool 🤩 I haven't updated any of my CI since I've been using mamba so I'm excited to see this in action.

jennyfothergill (approved):
LGTM! We must be on the same wavelength because I was just starting working on this. 😂

justinGilmer:
I had to add in a bit of support to locate the full path of the xmls for these tests. Not sure if this will solve them all, but will keep an eye out

.github/workflows/main.yml
justinGilmer: Big fan of this! If we need to add in other files as time goes on, we can

justinGilmer:
Tests are passing! Going to merge and we can add additional fixes/additions as needed.

PR #46, opened 2021-08-12
Add methods to construct system from job statepoints

daico007:
Add central method to construct `mb.Compound` from job statepoint. Still WIP (waiting for lists of params to be included in job statepoint to be finalized).


project/src/molecules/constructor.py
```diff
@@ -0,0 +1,33 @@
...
+    liq_box = mb.Box()
+
+    vap_box_cpd = mb.fill_box(
+        compound=molecule, n_compound=job.sp["n_compounds"], box=vap_box
+    )
+    liq_box_cpd = mb.fill_box(
+        compound=molecule, n_compound=job.sp["n_compounds"], box=liq_box
+    )
+
+    system = mb.Compound([vap_box_cpd, vap_box_cpd])
```
jennyfothergill, line 32:
```suggestion
    system = mb.Compound([liq_box_cpd, vap_box_cpd])
```
should one of these be the liquid box?
Also how will this handle if vap_box is None?

project/src/molecules/constructor.py
```diff
@@ -0,0 +1,33 @@
...
+    molecule = molecule_dict[job.sp["molecule"]]
+    vap_box = mb.Box()
+    liq_box = mb.Box()
+
+    vap_box_cpd = mb.fill_box(
+        compound=molecule, n_compound=job.sp["n_compounds"], box=vap_box
+    )
+    liq_box_cpd = mb.fill_box(
+        compound=molecule, n_compound=job.sp["n_compounds"], box=liq_box
+    )
```
jennyfothergill, line 30: Would "n_compounds" be split between the two boxes?

project/src/molecules/constructor.py
```diff
@@ -0,0 +1,33 @@
...
+
+    system = mb.Compound([vap_box_cpd, vap_box_cpd])
+    return system
```
jennyfothergill, line 33: Should we apply the forcefield before returning the system as well?

jennyfothergill (commented):
Nice work :) I have a couple questions/suggestions.

project/src/molecules/constructor.py
justinGilmer: There are now separate N_liq and N_vap variables in the job statepoints. If vapor phase is not needed, it is Null in the project statepoint. I think signac accesses that as `None`

project/src/molecules/constructor.py
daico007: Good catch! Fixed.
At the moment, I used a `if` to check if the `vap_box` is required.

project/src/molecules/constructor.py
daico007: This info is split into 2 in the new job statepoint set up.

project/src/molecules/constructor.py
daico007: I think this should return a mbuild compound, and the atomtyping step should happen in the individual `project.py`. I don't want to return a parmed structure here since some of our engines is requiring non-parmed writer.

justinGilmer: I think that should be a separate step, since we can serialize mbuild compounds to Json now. I have been burnt in the past where I have the initialize system including all these steps, but when FF application for more complicated systems can take a bit of time. I’d prefer to have the compound to inspect before we attempt to parametrize. We can make that a separate step that depends on the compound.Json files if we want. But I’m open for other suggestions.

project/src/molecules/system_builder.py
```diff
@@ -1,33 +1,65 @@
...
+    """
+    system = mb.Compound()
+
+    # Update this dict as new recipes are made
     molecule_dict = {
         "methaneUA": MethaneUA,
-        "pentaneUA": None,
+        "pentaneUA": PentaneUA,
         "benzeneUA": None,
         "waterSPC/E": None,
```
justinGilmer, line 44: Water spce already exists in mbuild itself, If you look at the test_spce.py file, I include the way to access that water model in there.

project/src/molecules/system_builder.py
```diff
@@ -1,33 +1,65 @@
...
     )
+    system.add(liq_box)
+
+    if sp["box_L_vap"] and sp["N_vap"]:
+        vap_box = mb.Box([sp["box_L_vap"]] * 3)
+        filled_vap_box = mb.fill_box(
+            compound=[molecule], n_compounds=[sp["N_vap"]], box=vap_box
+        )
+
+        # Translate the vapor box to be on top of the liquid box
```
justinGilmer, line 60: We need to verify with @ramanishsingh @rwsmith7531 and @bc118  if this is how they need the system to be set up. They might need two separate boxes anyway for their simulations.

justinGilmer (changes requested):
Small changes and then this should be good! Thanks @daico007 !

project/src/molecules/system_builder.py
daico007: gotcha, will add that in now

project/src/molecules/system_builder.py
daico007: Yes, I need some input about how the two box need to be stacked. Also, the individual box objects should still be accessible from the system (e.g., `system.children[0].box`)

ramanishsingh: Can we also return the molecule? Something like

```python
       return molecule, system
```
Or will it be messy?

Otherwise I can get the molecule from the recipe.

ramanishsingh: Also, right now the system is an mBuild compound. 

Instead, can we make a list of boxes? 

```python
    box_list=[]

    boxes.append(filled_liq_box)
    
    boxes.append(filled_vap_box)
    
    return box_list
```

And then we can access box_list[1] when we are in GEMC. 
I think that way it will be simpler if we keep the two boxes separate.

daico007: this is doable but it will make the output kinda messy (and may cause confusion for engines that do not need these info). I think these information can technically be retrieved from the `system` object without too much difficulty. 

Getting molecule info:
```
particle = system[1] # This is valid in here since we only have one kind of molecule
```

Getting list of boxes 
```
box_list = [child.box for child in system.children]
```

justinGilmer: I think @ramanishsingh means being able to access the vap and liq box as a list of two compounds instead of one. 

So for cases where N_vap or any of the vap statepoint items are None, it’s just a None in the location where the filled vap box would be. 

I think the main thing is that each of these compounds (liq and vap) do not have to be thought of as a single compound, but two compounds? @ramanishsingh is that correct?

justinGilmer: Or if we add labels for each compound in the system “liquid” and “vap” or something when adding them together could also work I think.

ramanishsingh: "I think the main thing is that each of these compounds (liq and vap) do not have to be thought of as a single compound, but two compounds? @ramanishsingh is that correct?"

Yes!

daico007: ok, I see. Then, how about we will return a single filled liquid box if the the vap box is None. If both of them exist, we will return a list of compound ([liq, vap]).

daico007: actually, would you prefer a list or a dictionary of system (i.e., `[liq_box, vap_box]` vs `{"liquid": liq_box, "vapor": vap_box}`?

daico007: Fix included in the latest commit (right now I am returning a list of boxes, but lmk if you prefer a dictionary)

ramanishsingh: Thanks!

project/src/molecules/system_builder.py
```diff
@@ -1,33 +1,59 @@
...
-    system = mb.Compound([vap_box_cpd, vap_box_cpd])
-    return system
+    if sp["box_L_vap"] and sp["N_vap"]:
+        vap_box = mb.Box([sp["box_L_vap"]] * 3)
+        filled_vap_box = mb.fill_box(
+            compound=[molecule], n_compounds=[sp["N_vap"]], box=vap_box
+        )
+        return [filled_liq_box, filled_vap_box]
+    else:
+        return filled_liq_box
```
bc118, line 59: I think we should have the same length list here with the with vapor box as None : 

return [filled_liq_box, filled_vap_box]
     else:
         return  [filled_liq_box, None]

bc118 (changes requested):
I am not a GROMACS user, so I have no input on any of the GROMCAS files.  I think the system builder is OK aside from the one comment, which I believe just makes it more flexible, having the same list length for all outputs.

project/src/molecules/system_builder.py
daico007: done

PR #49, opened 2021-08-16
Overhaul statepoint generation, addl. info added

justinGilmer:
Previously, we were using a standard combinatoric product to generate
all the possible statepoints for the project.

This would work correctly, if each parameter we are changing applies to
every system. However, this is not the case. This overhaul attempts to
provide a bit more of an extensible setup for each molecule, as well as
being a bit more tunable based on the specific statepoint information.

### DIFF
--- a/project/init.py
+++ b/project/init.py
@@ -3,6 +3,8 @@ import itertools
 import os
 
 import signac
+import unyt as u
+from numpy import ModuleDeprecationWarning
 
 
 def dict_product(dd):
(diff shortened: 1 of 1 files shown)


justinGilmer:
This also includes `unyt` for the parameters and conversions are done during statepoint creation. Units are not written out to the json files, but if we are certain we are writing out to the correct units at first, then we can load these in with confidence.

daico007 (approved):
LGTM!

PR #52, opened 2021-08-16
Add Lammps Workflow

CalCraven:
Setup a project.py signac workflow that will analyze statepoints and create trajectories from lammpsdata output

- [x] Create a list of labels for flow progress
- [x] Initialize molecules from statepoint
- [x] Copy over proper run files
- [x] run lammps script on PBS or SLURM
- [x] Analyze trajectory info for density and rdf output


justinGilmer (changes requested):
Small change I noticed, overall, looking good!

justinGilmer:
Pre-commit is failing, which means it is a lint that is not able to be automatically applied, most likely missing doctrings. You will need to install pre-commit locally and run it to identify those.

https://github.com/mosdef-hub/reproducibility_study/blob/main/CONTRIBUTING.md#using-pre-commit-to-auto-format-commits

reproducibility_project/src/engines/lammps-UD/project.py
```diff
@@ -0,0 +1,205 @@
...
+    elif job.sp.forcefield_name == "oplsaa":
+        ff = foyer.Forcefield(name="oplsaa")
+    elif job.sp.forcefield_name == "spce":
+        ff = foyer.Forcefield(
+            name="spce"
+        )  # TODO: Make sure this gets applied correctly
+    else:
+        raise Exception(
+            "No forcefield has been applied to this system {}".format(job.id)
+        )
```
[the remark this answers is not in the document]
CalCraven: I think the plan is for UD to copy the methods that use the FF utility in lammps-VU/project.py. For now, we may merge the current workflow, and then @zijiewu3 will pull his own PR to get his project.py file going.

reproducibility_project/src/engines/lammps-VU/project.py
```diff
@@ -0,0 +1,237 @@
...
+def lammps_calc_rdf(job):
+    """Create an rdf from the gsd file using Freud analysis scripts."""
+    # Create rdf data from the production run
+    import mbuild as mb
+    import mdtraj
+
+    traj = md.load("prod.xtc", top="box.gro")
+    traj.save("prod.gsd")
+    # TODO: Use freud rdf PR to create an RDF from the gsd file
+    return
```
justinGilmer, line 222: Just a heads up, this should not be handled here in this project.py. That will be handled solely by the `project-analysis.py` file located with the `init.py, src/ ` etc.

So instead you should just need a method to convert your prod traj to `trajectory.gsd`

justinGilmer (approved):
Just a small comment, but lgtm!

reproducibility_project/src/engines/lammps-VU/project.py
CalCraven: Noted! We should be good to go then.

PR #54, opened 2021-08-17
Sanitize hoomd outputs

jennyfothergill:
- [x] add logfile
- [x] convert logged quantities to correct units
- [x] check units on gsdfile


jennyfothergill:
I'm pretty sure (although I'd like a second opinion) that with the conversions I applied to the hoomd3_forcefield, all the units in the output should be correct.

project/src/engines/hoomd/project.py
```diff
@@ -32,34 +32,60 @@ def run_hoomd(job):
...
 
     write_gsd(structure, job.fn("init.gsd"), ref_distance=rd, ref_energy=re)
-    # NOTE: create_hoomd3_forcefield depends on PR
-    # https://github.com/mosdef-hub/mbuild/pull/871
+    # ref_distance: angstrom -> nm
+    # ref_energy: kcal/mol -> kJ/mol
+    # ref_mass: dalton -> amu
     snapshot, forcefield, ref_vals = create_hoomd3_forcefield(
-        structure, ref_distance=10, ref_energy=1 / 4.184
+        structure, ref_distance=0.1, ref_energy=1 / 4.184, ref_mass=0.9999938574
```
tcmoore3, line 39: The units in mBuild + foyer are already in nm, and the energies are already in kJ/mol. Do we need to do these conversions here?

project/src/engines/hoomd/project.py
```diff
@@ -32,34 +32,60 @@ def run_hoomd(job):
...
+        thermo_props,
+        quantities=[
+            "kinetic_energy",
+            "potential_energy",
+            "pressure",
+            "kinetic_temperature",
+            "volume",
+        ],
+    )
+    file = open("log.txt", mode="x", newline="\n")
```
tcmoore3, line 68: This mode will throw an error if the file already exists, which may happen if we restart the simulation, e.g., to let it run longer.

project/src/engines/hoomd/project.py
```diff
@@ -32,34 +32,60 @@ def run_hoomd(job):
...
+    table_file = hoomd.write.Table(
```
tcmoore3, line 69: Was there a strong preference for the table file over a gsd to log values?

project/src/engines/hoomd/project.py
```diff
@@ -32,34 +32,60 @@ def run_hoomd(job):
...
+        trigger=hoomd.trigger.Periodic(period=5000),
+        logger=logger,
+        max_header_len=7,
+    )
+    sim.operations.writers.append(table_file)
+
     integrator = hoomd.md.Integrator(dt=0.005)
     integrator.forces = forcefield
     nvt = hoomd.md.metehods.NVT(
         filter=hoomd.filter.All(), kT=job.sp.temperature, tau=1.0
```
tcmoore3, line 80: We may have to convert the temperatures here into energy units.

project/src/engines/hoomd/project.py
tcmoore3: If we're not totally set on table files, I would vote for GSD files because:
1. They are _not_ human-readable, which forces you to process them programmatically 
2. You don't really have to worry about headers and stuff, the values are stored in the file with descriptive field names
3. They are much more space-efficient than text files (although that's probably not an issue for this particular project)

project/src/engines/hoomd/project.py
```diff
@@ -32,34 +32,60 @@ def run_hoomd(job):
...
+    )
+    sim.operations.computes.append(thermo_props)
+    logger.add(
+        thermo_props,
+        quantities=[
+            "kinetic_energy",
+            "potential_energy",
+            "pressure",
+            "kinetic_temperature",
+            "volume",
```
tcmoore3, line 65: What about the bond, angle, etc... energies?

project/src/engines/hoomd/project.py
jennyfothergill: yes, I think so  because the parmed structure expects [angstrom, kcal, and dalton](https://github.com/mosdef-hub/mbuild/blob/master/mbuild/formats/hoomd_forcefield.py#L68-L72)

project/src/engines/hoomd/project.py
jennyfothergill: good point! I could change mode to `'a'` to fix this?

project/src/engines/hoomd/project.py
jennyfothergill: I am going off [this reference card](https://github.com/mosdef-hub/reproducibility_study/projects/1#card-66677615) from the project board. We chose not to combine them in one file since the write out periods would not be the same. I don't have a strong preference, but it is nice that the values are human readable.

project/src/engines/hoomd/project.py
jennyfothergill: I can add those, but they may not be comparable across engines.

project/src/engines/hoomd/project.py
tcmoore3: OK. So you pass the value you divide by to get from the parmed units to the nm/kJ/amu units. So `ref_distance` should be 10 and `ref_energy` should be 1/4.184

project/src/engines/hoomd/project.py
jennyfothergill: good point!

project/src/engines/hoomd/project.py
tcmoore3: That should work

project/src/engines/hoomd/project.py
tcmoore3: We can write to multiple GSD files, one for the trajectory and one more frequently for the thermo data without trajectory data.

project/src/engines/hoomd/project.py
tcmoore3: I guess the bond energies will be 0 since they are rigid (same with angles in water?). But it's still a good idea to have them available even if we end up not using that data.

project/src/engines/hoomd/project.py
tcmoore3: All that being said, let's not let this hold us up for actually generating the data. I'm OK with text files.

project/src/engines/hoomd/project.py
jennyfothergill: 🤦‍♀️ Yes, you're right. Sorry I realize this is how you had it and I just got confused. I've fixed it and updated the note above to make it more clear to myself!

project/src/engines/hoomd/project.py
justinGilmer: Unyt should be able to handle that for us. https://unyt.readthedocs.io/en/stable/modules/unyt.equivalencies.html#unyt.equivalencies.ThermalEquivalence


```python3
import unyt as u
temp = 300 * u.K
energy_temp = temp.to_equivalent('eV', 'thermal')
```

In this case, I get for 300K, a thermal energy equivalent to ~0.025 eV which is what [wikipedia reports in the "Temperature" section](https://en.wikipedia.org/wiki/Electronvolt)

project/src/engines/hoomd/project.py
justinGilmer: Since it will be a bit more of a challenge to convert all the other engine's thermodynamic data to a GSD compared to just the trajectory, I am a bit more in favor of our engines all writing out to something a bit more similar for this case.

project/src/engines/hoomd/project.py
justinGilmer: Not entirely sure what else is needed here to make it unitless like hoomd expects? @tcmoore3 or is converting it to energy units sufficient?

jennyfothergill: That's awesome! In this case, we'd want to convert to kJ/mol, right?

justinGilmer (commented):
Looks good. I think once we are certain how we need to convert the temperature into energy units, this should now be g2g!

project/src/engines/hoomd/project.py
jennyfothergill: @justinGilmer I think if all the units (distance/mass/energy) are converted by the `create_hoomd_forcefield`, keeping the kT energy value consistent should do it.

tcmoore3: Yes, I think that unyt code is what we want. In this case, it would be something like:
```python
>>> import unyt as u
>>> temp = 300 * u.K
>>> kT = temp.to_equivalent('kJ/mol', 'thermal')
>>> kT
unyt_quantity(2.49433864, 'kJ/mol')
```

justinGilmer: Perfect! Thanks for testing that too!

project/src/engines/hoomd/project.py
```diff
@@ -32,34 +33,60 @@ def run_hoomd(job):
...
+    )
+    sim.operations.writers.append(table_file)
+
     integrator = hoomd.md.Integrator(dt=0.005)
     integrator.forces = forcefield
-    nvt = hoomd.md.metehods.NVT(
-        filter=hoomd.filter.All(), kT=job.sp.temperature, tau=1.0
-    )
+    # convert temp in K to kJ/mol
+    kT = float((300 * u.K).to_equivalent("kJ/mol", "thermal"))
```
justinGilmer, line 81: This should be job.sp.temperature instead of 300

justinGilmer (changes requested):

project/src/engines/hoomd/project.py
justinGilmer: Also if you just want the value returned without the units attached you can do:

```suggestion
    kT = (300 * u.K).to_equivalent("kJ/mol", "thermal").value
```

jennyfothergill: > This should be job.sp.temperature instead of 300

oops, yep!

jennyfothergill: > Also if you just want the value returned without the units attached you can do:

doesn't this return an array? I'm not sure what hoomd will do with that.

project/src/engines/hoomd/project.py
justinGilmer: I agree.

project/src/engines/hoomd/project.py
justinGilmer: In cases where the unyt is applied to a list-like structure, it returns an array, but in the case where it is a single value, it behaves like one.

justinGilmer (approved):

PR #55, opened 2021-08-17
Add equilibration sampling step

chrisjonesBSU:
This PR adds an equilibration sampling step that uses the Pymbar timeseries module https://github.com/choderalab/pymbar, https://pymbar.readthedocs.io/en/latest/timeseries.html

It will write out the beginning and end indices of the production region, and the stride/step size that corresponds to uncorrelated samples to the job document. 

It defaults to using potential energy for identifying production region and sub sampling, but this is a parameter in the function that can be changed. 

As a quick note, I set `conservative = True` in the `subsampleCorrelatedData` function call. It makes things a bit easier for us. From the docs:

> conservative : bool, optional, default=False
> if set to True, uniformly-spaced indices are chosen with interval ceil(g), 
> where g is the statistical inefficiency. Otherwise, indices are chosen 
> non-uniformly with interval of approximately g in order to end up with
> approximately T/g total indices


project/src/analysis/sampler.py
```diff
@@ -0,0 +1,27 @@
+"""Use the pymbar package to perform decorrelated equilibration sampling"""
+
+import numpy as np
+from pymbar import timeseries
+
+
+def decorr_sampling(job, data_type="potential_energy"):
```
jennyfothergill, line 7: we might break this into two functions to make testing easier. e.g., one function handles the job and the other just wraps the equilibration/decorrelation. ([Here's an example of what I mean from the rdf PR](https://github.com/mosdef-hub/reproducibility_study/pull/35/commits/39860a633ff97a8e177980fb82b9abec14cd554c) thanks to a suggestion from @justinGilmer)

project/src/analysis/sampler.py
```diff
@@ -0,0 +1,27 @@
+"""Use the pymbar package to perform decorrelated equilibration sampling"""
```
jennyfothergill, line 1:
```suggestion
"""Use the pymbar package to perform decorrelated equilibration sampling."""
```

project/src/analysis/sampler.py
```diff
@@ -0,0 +1,27 @@
...
+    """Uses the timeseries module from pymbar to perform statistical sampling.
+
+    The start, end and decorrleated step size of the production region are
+    added to the job document.
+
+    Parameters
+    ----------
+    job : signac.contrib.job.Job
+        The Job object.
+    data_type : str; default "potential_erngy"
```
jennyfothergill, line 17:
```suggestion
    data_type : str; default "potential_energy"
```

jennyfothergill (commented):
Nice work! I have a couple of suggestions and also think we should add a unit test to check that this is working correctly.

justinGilmer:
I pushed some additional edits to begin to incorporate the helper functions from #56 as well as fixed some pre-commit errors.

project/src/analysis/sampler.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    data = np.genfromtxt(job.fn("log.txt", names=True))[data_type]
+    is_equil, prod_start, ineff = is_equilibrated(data, threshold=0.8, nskip=1)
+    if is_equil:
+        uncorr_indices = timeseries.subsampleCorrelatedData(
+            data[prod_start:], g=ineff, conservative=True
+        )
+        job.doc["sample_start"] = uncorr_indices.start + prod_start
+        job.doc["sample_end"] = uncorr_indices.stop + prod_start
+        job.doc["sample_stride"] = uncorr_indices.step
```
justinGilmer, line 30: Based on https://pymbar.readthedocs.io/en/master/timeseries.html#pymbar.timeseries.subsampleCorrelatedData

I am not sure where the uncorr_indices.start, stop, step come from? `subsampleCorrelatedData` does not seem to return that kind of information?

justinGilmer (changes requested):
If you can answer the question I have below, once we have some unit tests added and modify the function to resemble the RDF setup that @jennyfothergill recommended, this should be good to go!

project/src/analysis/sampler.py
jennyfothergill: `subsampleCorrelatedData` returns a range--not a list as said in the docs. Anyway, this allows us to access the start, stop, and step info.

chrisjonesBSU: It depends if `conservative` is set to `True` or `False`. When `True` it returns a `range`, when it's `False` it will return a list

project/src/analysis/sampler.py
chrisjonesBSU: Ok, this is done. Let me know if this is what you had in mind.

jennyfothergill (approved):
LGTM

project/src/analysis/sampler.py
jennyfothergill: Ah OK, I didn't realize that.

project/src/analysis/sampler.py
jennyfothergill: LGTM c:

reproducibility_project/src/analysis/sampler.py
```diff
@@ -0,0 +1,57 @@
...
+    job : signac.contrib.job.Job
+        The Job object.
+    variable : str; default "potential_energy"
+        The variable to be used in sampling.
+    threshold : float, optional, default=0.75
+        Fraction of data expected to be equilibrated.
+    """
+    data = np.genfromtxt(job.fn("log.txt"), names=True)[variable]
+    start, stop, step = _decorr_sampling(data, threshold)
+    job.doc["sampled_variable"] = variable
```
justinGilmer, line 26: If we want to measure two properties like "density" and "potential energy", I dont think this would be able to store the `sample_start`, etc. for multiple properties. We might need to adjust the variable names or maybe write out a separate mini json files for each property.

justinGilmer (changes requested):
Small change to the variable name I noticed.

reproducibility_project/src/analysis/sampler.py
chrisjonesBSU: If we want to sample multiple variables and store the params for each, maybe a dictionary would work?

```
sample_results = {}
sample_results[variable] = range(start, stop, step)

job.doc["sample_results"] = sample_results:
```

chrisjonesBSU: @justinGilmer I just pushed a commit along the lines of my last comment, let me know what you think of this approach.

reproducibility_project/src/analysis/sampler.py
```diff
@@ -0,0 +1,59 @@
...
+    ----------
+    job : signac.contrib.job.Job
+        The Job object.
+    variable : str; default "potential_energy"
+        The variable to be used in sampling.
+    threshold : float, optional, default=0.75
+        Fraction of data expected to be equilibrated.
+    """
+    try:
+        job.doc["sample_results"]
```
justinGilmer, line 25: Lets name it `sampling_results` to be a bit more descriptive, but other  than that, it looks great!

project/src/analysis/sampler.py
justinGilmer: Great catch, I missed that and I assumed that `range` would return a list, but that does not make sense since it is a range object.

justinGilmer (approved):
Thanks for this, @chrisjonesBSU ! Once the macos build passes, this can be merged!

PR #56, opened 2021-08-17
Add helper methods for timeseries support

justinGilmer:
This adds in helper functions to detect equilibration in timeseries
data, as well as trimming away the timeseries data that is not
equilibrated.

Unit tests are included to verify behavior.

### DIFF
--- a/project/src/analysis/equlibration.py
+++ b/project/src/analysis/equlibration.py
@@ -1,0 +1,85 @@
+"""Timeseries and pyMBAR related methods."""
+from typing import List
+
+import numpy as np
+import numpy.typing as npt
+from pymbar import timeseries
+
+
+def is_equilibrated(
+    a_t: npt.ArrayLike, threshold: float = 0.8, nskip: int = 1
+) -> bool:
+    """Check if a dataset is equilibrated based on a fraction of equil data.
+
+    Using `pymbar.timeseries` module, check if a timeseries dataset has enough
+    equilibrated data based on a threshold value. The threshold value
+    translates to the fraction of total data from the dataset 'a_t' that
+    can be thought of as being in the 'production' region.
+
+    The `pymbar.timeseries` module returns the starting index of the
+    'production' region from 'a_t'. The fraction of 'production' data is
+    then compared to the threshold value. If the fraction of 'production' data
+    is >= threshold fraction this will return True, False otherwise.
+
+    Parameters
+    ----------
+    a_t : numpy.typing.Arraylike
+        1-D time dependent data to check for equilibration.
+    threshold : float, optional, default=0.8
+        Fraction of data expected to be equilibrated.
+    nskip : int, optional, default=1
+        Since the statistical inefficiency is computed for every time origin
+        in a call to timeseries.detectEquilibration, for larger datasets
+        (> few hundred), increasing nskip might speed this up, while
+        discarding more data.
+    """
+    if threshold < 0.0 or threshold > 1.0:
+        raise ValueError(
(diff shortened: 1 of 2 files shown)


justinGilmer:
@chrisjonesBSU can you take a look at this and see if we can leverage utilities like these in your PR? This should provide a defined method we can all use in the study to then sample data without much code duplication for different properties.

project/src/analysis/equlibration.py
```diff
@@ -0,0 +1,85 @@
...
+    else:
+        return False
+
+
+def trim_non_equilibrated(
+    a_t: npt.ArrayLike, threshold: float = 0.75, nskip: int = 1
+) -> List:
+    """Prune timeseries array to just the production data.
+
+    Refer to equilibration.is_equilibrated for addtional information.#!/usr/bin/env python
```
jennyfothergill, line 55:
```suggestion
    Refer to equilibration.is_equilibrated for addtional information.
```

project/tests/test_equilbration.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def test_incorrect_threshold(self):
+        data = testsystems.correlated_timeseries_example(
+            N=1000, tau=200, seed=432
+        )
+        with pytest.raises(ValueError, match=r"Passed \'threshold\' value"):
+            is_equilibrated(data, threshold=2.0)
+
+        with pytest.raises(ValueError, match=r"Passed \'threshold\' value"):
+            is_equilibrated(data, threshold=-2.0)
```
jennyfothergill, line 30: is one of these supposed to be negative?

jennyfothergill (approved):

project/src/analysis/equlibration.py
```diff
@@ -0,0 +1,85 @@
...
+    threshold : float, optional, default=0.8
+        Fraction of data expected to be equilibrated.
+    nskip : int, optional, default=1
+        Since the statistical inefficiency is computed for every time origin
+        in a call to timeseries.detectEquilibration, for larger datasets
+        (> few hundred), increasing nskip might speed this up, while
+        discarding more data.
+
+    """
+    if not is_equilibrated(a_t, threshold=threshold, nskip=nskip):
```
chrisjonesBSU, line 78: When calling the `is_equilibrated` function, if it's True, could it also return `t0` and `g`?  That way we don't have to run `timeseries.detectEquilibration` a second time in `trim_non_equilibrated`

project/tests/test_equilbration.py
justinGilmer: Yeah I just wanted to make sure my validation check for `threshold` was behaving as expected. Only allow floats between 0.0-1.0

project/src/analysis/equlibration.py
justinGilmer: I haven’t tried that in python, but would that be an issue for the false case? 

Or I guess we can just return None for those values. Does `None` for the non-existent t0 and g seem reasonable in the false case?

project/src/analysis/equlibration.py
justinGilmer: Good catch, my ide autocomplete thought it was being helpful with the tab completion 😂

project/src/analysis/equlibration.py
jennyfothergill: yeah, `None` seems reasonable-- in this case we would use it something like this?
```
eq, t0, g = is_equilibrated
if eq:
    do something
```

justinGilmer: Ok cool. I just pushed those changes!

PR #59, opened 2021-08-17
Patch for `construct_system`

daico007:
Fix bug and add tests for `construct_system`


daico007:
Still working on adding more details test

daico007:
@ramanishsingh Can you try if this fix your issue?

project/src/molecules/system_builder.py
```diff
@@ -1,8 +1,9 @@
 """Methods used to create systems from job statepoint."""
 import mbuild as mb
 from mbuild.lib.molecules.water import WaterSPC
-from methane_ua import MethaneUA
-from pentane_ua import PentaneUA
+
+from project.src.molecules.methane_ua import MethaneUA
```
ramanishsingh, line 5: I get this error

ModuleNotFoundError: No module named 'project.src'; 'project' is not a package

ramanishsingh: This works for me:

```python
from methane_ua import MethaneUA
from pentane_ua import PentaneUA
```

justinGilmer: Where were you running this from @ramanishsingh ? This should be ran inside the `project` folder.

```bash
conda activate mosdef-study38
cd reproducibility_study/project
python src/engines/mcccs/project.py
```

At least I am pretty sure that is the case

project/src/molecules/system_builder.py
```diff
@@ -37,13 +38,15 @@ def construct_system(sp):
...
-        "pentaneUA": PentaneUA,
+        "methaneUA": MethaneUA(),
+        "pentaneUA": PentaneUA(),
         "benzeneUA": None,
-        "waterSPC/E": WaterSPC,
+        "waterSPC/E": WaterSPC(),
         "ethanolAA": None,
     }
     molecule = molecule_dict[sp["molecule"]]
+    # print(molecule)
```
justinGilmer, line 48: Once we are happy with this, we should remove this print statement

project/src/molecules/system_builder.py
ramanishsingh: Is it working for you?
I am still getting project is not a package error.

daico007: I changed both relative statement, waiting for tests.

daico007: I revert to the original import statement since `from pentane_ua import PentaneUA` would not pass the CI tests

daico007: I think this import issue has something to do with the library is named `project` get imported in a directory that also has a `project.py`. I am not sure if there is an easy fix to this, but we may consider changing the library name (maybe `reproducibility_project`)? @justinGilmer

justinGilmer: Ahh. Yeah that might be it @daico007     If you want to try changing the library name to reproducibility_project, let’s see if that fixes it

daico007: @justinGilmer I changed the name of the main dir to be `reproducibility_project`. @ramanishsingh can you try and see if this works for you.

ramanishsingh: This worked for me after I added this in my `project.py`

```python
sys.path.append(Project().root_directory() + "/..")
```

justinGilmer (approved):

PR #60, opened 2021-08-18
Update hoomd project to work with new init and construct system

jennyfothergill:
I think at this point we should have everything we need to run. This an attempt to find/fix any stumbling points.

- [ ] check how to update restart logic. We used to use [HOOMD_WALLTIME_STOP env variable](https://hoomd-blue.readthedocs.io/en/stable/restartable-jobs.html#cleanly-exit-before-the-walltime-limit), does it still work?


reproducibility_project/src/engines/hoomd/project.py
```diff
@@ -20,24 +20,56 @@ def __init__(self):
...
+    # temporary hack until benzene and ethanol are added
+    try:
+        # Ignore the vapor box
+        filled_box, _ = construct_system(job.sp)
+    except AttributeError:
+        return
+
+    if job.sp.forcefield_name == "trappe-ua":
+        ff = foyer.forcefields.load_TRAPPE_UA()
+    elif job.sp.forcefield_name == "benzene_ua":
```
justinGilmer, line 46: This name was also updated in #62 
```suggestion
    elif job.sp.forcefield_name == "benzene-ua":
```

justinGilmer (changes requested):
one small change so far

jennyfothergill:
This is currently running on our lab's cluster :)

reproducibility_project/src/engines/hoomd/project.py
```diff
@@ -83,7 +133,28 @@ def run_hoomd(job):
...
+    final_box = hoomd.Box(Lx=L, Ly=L, Lz=L)
+    box_resize_trigger = hoomd.trigger.Periodic(10)
+    box_resize = hoomd.update.BoxResize(
+        box1=initial_box,
+        box2=final_box,
+        variant=ramp,
+        trigger=box_resize_trigger,
+    )
+    sim.operations.updaters.append(box_resize)
+    sim.run(2e4 + 1)
```
justinGilmer, line 152: Just for my own curiosity, what is the reasoning for 2e4 + 1?

jennyfothergill: 2e4 is the box shrink time, by running one more step the box is fully shrunk and the following assert line will evaluate to true

justinGilmer: Ah gotcha! I missed that.

justinGilmer (approved):

PR #62, opened 2021-08-18
Tidy init.py by using itertools.product

justinGilmer:
Previously we used a naive implementation of itertools.product to
generate all possible statepoints to examine.

However, there were concerns with that setup since a lot of our
statepoint variables depended on other statepoint variables, so a simple
product would not work properly. This was remedied by using dictionaries
to store this info based on other statepoint keys.

As @jennyfothergill pointed out, we can now use itertools.product again,
since the dictionary representation all use lists which are varaible
length based on the other statepoint information. Namely, the molecule
name and the simulation engine.

### DIFF
--- a/reproducibility_project/init.py
+++ b/reproducibility_project/init.py
@@ -163,10 +163,7 @@ for molecule in molecules:
                 temp.to_value("K"),
                 decimals=3,
             ),
-            "pressure": np.round(
-                press.to_value("kPa"),
-                decimals=3,
-            ),
+            "pressure": np.round(press.to_value("kPa"), decimals=3).item(),
             "ensemble": ensemble if ensemble else None,
             "N_liquid": n_liq,
             "N_vap": n_vap if n_vap else None,
--- a/reproducibility_project/init.py
+++ b/reproducibility_project/init.py
@@ -36,7 +36,7 @@ for key in molecules:
         if "benz" not in key:
             forcefields[key] = "trappe-ua"
         else:
-            forcefields[key] = "benzene_ua"
+            forcefields[key] = "benzene-ua"
         r_cuts[key] = 14.0 * u.angstrom
     elif "SPC/E" in key:
         forcefields[key] = "spce"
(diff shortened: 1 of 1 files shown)


daico007:
when you're at this, can you also change the `"Trappe-UA"` to `"trappe-ua"` (match with the name used in foyer Forcefield)? it makes the loading forcefield from by name a bit easier

jennyfothergill (approved):
Thank you! That's a little easier to read, for me at least. ☺️ 
One note that I don't know if we need to do anything about is that using unyt quantities (as they are arrays even if they're just one value) makes signac throw this warning 
```
NumpyConversionWarning: Any numpy types provided will be transparently converted to the closest base Python equivalents.
```
which I think is basically saying that if we pass a 0D array it will return the item in the array
https://github.com/glotzerlab/signac/blob/e60bfbc579371698c2690605d10f88b3a12bc27c/signac/synced_collections/numpy_utils.py

reproducibility_project/init.py
```diff
@@ -130,87 +130,80 @@ def dict_product(dd):
...
+            "engine": engine,
+            "replica": replica,
+            "temperature": np.round(
+                temp.to_value("K"),
+                decimals=3,
+            ),
+            "pressure": np.round(
+                press.to_value("kPa"),
+                decimals=3,
+            ),
```
jennyfothergill, line 169:
```suggestion
            "pressure": np.round(press.to_value("kPa"), decimals=3).item(),
```
if we want to silence the signac/numpy warning, we could do this

jennyfothergill:
Oh, and another thought--in case we want to use 
```
signac view
```
we should change the statepoint `waterSPC/E` to `waterSPCE`. Otherwise we'll get this error:
```
Error: In order to use view, statepoints should not contain ['/', ' ', '*']
```

justinGilmer:
Great catches @daico007 and @jennyfothergill !

I will make those now

justinGilmer:
Ok I think i fixed all the issues. When i re-generated the statepoints, i had no warnings!

daico007 (approved):

PR #63, opened 2021-08-18
Refining GROMACS script

daico007:
WIP - Refining GROMACS workflow scripts (input conditions + `project.py`)

### DIFF
--- a/reproducibility_project/src/engines/gromacs/project.py
+++ b/reproducibility_project/src/engines/gromacs/project.py
@@ -1,12 +1,9 @@
 """Setup for signac, signac-flow, signac-dashboard for this study."""
 import os
 import pathlib
+import sys
 
 import flow
-from flow import environments
-
-import reproducibility_project
-from reproducibility_project.src.engine_input.gromacs import mdp
 
 
 class Project(flow.FlowProject):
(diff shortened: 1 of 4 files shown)


reproducibility_project/src/engines/gromacs/project.py
```diff
@@ -19,171 +16,138 @@ def __init__(self):
...
+            "fname": "nvt.mdp",
+            "template": f"{mdp_abs_path}/nvt_template.mdp.jinja",
+            "data": {"temp": job.sp.temperature},
+        },
+        "npt": {
+            "fname": "npt.mdp",
+            "template": f"{mdp_abs_path}/npt_template.mdp.jinja",
+            "data": {
+                "temp": job.sp.temperature,
+                "refp": job.sp.pressure * 0.01,  # convert from kPa to bar
```
justinGilmer, line 67: Lets leverage unyt here if we can

```py3
import unyt as u
press = job.sp.pressure * u.kPa
                "refp": press.to_value('bar'),  # convert from kPa to bar
```

justinGilmer (changes requested):
Small changes, but this looks good!

reproducibility_project/src/engines/gromacs/project.py
```diff
@@ -19,171 +16,174 @@ def __init__(self):
...
+        ff_name = "benzene_trappe-ua_like.xml"
+        ff_path = (
+            str(os.path.dirname(os.path.abspath(xmls.__file__))) + "/" + ff_name
+        )
+        system[0].save(
+            filename="init.top", forcefield_files=ff_path, overwrite=True
+        )
+
+    # Modify mdp files according to job statepoint parameters
+    import unyt as u
```
justinGilmer, line 65: Lets move this to the top of the file with the other imports.

justinGilmer (changes requested):
Small change, then LGTM!

reproducibility_project/src/engines/gromacs/project.py
daico007: Fixed

justinGilmer (approved):
Once we change the center of mass motion removal back from Angular to linear, this is good to go.

PR #70, opened 2021-08-20
Feat/utility forcefield loader

justinGilmer:
Previously, each project.py needed special logic to handle loading the
proper forcefield and xml.

This is no longer needed, as it can lead to a lot of unexpected issues
between implementations. A utils directory was created with the hope of
adding useful methods that most of the projects can use and hopefully
reduce the amount of duplicated code in the project.

The project.py files for gromacs, hoomd, and mcccs have been modified to
use this new method as well as unit tests.


jennyfothergill (approved):
LGTM!

PR #71, opened 2021-08-20
Add Cassandra project.py

rwsmith7531:


reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -17,6 +17,255 @@ def __init__(self):
...
+    ffname = job.sp.forcefield
+    ff = get_ff(ffname) # make this function exist
+    structure = ff.apply(compound)
+
+    T = job.sp.temperature * u.K
+    P = job.sp.pressure * u.kPa
+
+
+    species_list = [structure]
+    cutoff_dict = {
```
justinGilmer, line 75: This information is actually part of the statepoint for each job now.

`job.sp.r_cut`

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -17,6 +17,255 @@ def __init__(self):
...
+    N = Nliq+Nvap
+
+    filled_boxes = construct_system(job.sp)
+
+    liqbox_filled = filled_boxes[0]
+    if ensemble == "GEMC-NVT":
+        vapbox_filled = filled_boxes[1]
+
+    ffname = job.sp.forcefield
+    ff = get_ff(ffname) # make this function exist
```
justinGilmer, line 67: You can use the utils methods that should be merged soon. See #70 

```py3
from reproducibility_project.src.utils.forcefields import load_ff
ff = load_ff(job.sp.forcefield_name)
```

reproducibility_project/src/engines/cassandra/project.py
justinGilmer: Though these are in nm and need to be converted for your case

justinGilmer (changes requested):
Few small changes so far, but looks great @rwsmith7531 !

reproducibility_project/src/engines/cassandra/project.py
rwsmith7531: I fixed it and should push the changes soon.

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -17,6 +17,255 @@ def __init__(self):
         self.ff_fn = self.data_dir / "forcefield.xml"
 
 
+@Project.label
+def is_cassandra(job):
+    return job.sp.simulation_engine == "cassandra"
+
+
+@Project.operation
```
justinGilmer, line 25: This will need either the is_cassandra call above for each call, or as a lambda for each operation
```
@Project.operation
@Project.pre(lambda j: j.sp.engine == "cassandra")
@Project.post(cassandra_complete)
@directives(omp_num_threads=4)

File: reproducibility_project/src/engines/cassandra/project.py [hunk 4]
```diff
@@ -17,6 +17,255 @@ def __init__(self):
         self.ff_fn = self.data_dir / "forcefield.xml"
 
 
+@Project.label
+def is_cassandra(job):
+    return job.sp.simulation_engine == "cassandra"
```
justinGilmer, line 22:
```suggestion
    return job.sp.engine == "cassandra"
```

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -17,6 +17,255 @@ def __init__(self):
...
+    cbmc_n_ins = 12
+    cbmc_n_dihed = 50
+
+    prop_freq = 10
+    coord_freq = 10
+    if ff_name == "oplsaa":
+        comb_rule = "geometric"
+    else:
+        comb_rule = "lb"
+
```
justinGilmer, line 112: You can now access this programmatically if you want.

```suggestion
if ff.combining_rule == "geometric":
    comb_rule = "geometric"
else:
    comb_rule = "lb"
```

justinGilmer (changes requested):
Small changes I noticed. Current edits look good though!

justinGilmer:
Also, pre-commit is failing and is not able to automatically fix it and commit them. It is most likely docstring issues. You will need to install pre-commit locally to fix that. [How to do that is located here.](https://github.com/mosdef-hub/reproducibility_study/blob/main/CONTRIBUTING.md#using-pre-commit-to-auto-format-commits)

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -17,7 +19,363 @@ def __init__(self):
...
+    moveset = mc.MoveSet(cass_ensembles[ensemble], species_list)
+    moveset.prob_volume = p_volume
+    moveset.prob_translate = p_translate
+    moveset.prob_rotate = p_rotate
+    moveset.prob_regrow = p_regrow
+    moveset.cbmc_n_ins = cbmc_n_ins
+    moveset.cbmc_n_dihed = cbmc_n_dihed
+
+    with job:
```
justinGilmer, line 202: If you want, you can remove this context management statement and add the `@flow.with_job` decorator at the top with your other flow directives. This behaves the same way as the
```py3
with job:
    ...
``` 
context manager, but saves you a level of indentation

See [here](https://docs.signac.io/projects/flow/en/latest/api.html#flow-with-job)

reproducibility_project/templates/crc.nd.sh
```diff
@@ -0,0 +1,15 @@
...
+#$ -pe smp {{ np_global }}
+#$ -r n
+#$ -q long
+#$ -m abe
+#$ -M rsmith56@nd.edu
+
+conda activate mosdef-study38
+
+{% block tasks %}
+{% endblock %}
```
justinGilmer, line 14: Probably not too important, but do you want to start the `block tasks` before you close the `block header`?

If not, i would move one of the `endblock` commands to line 12

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -1,10 +1,12 @@
...
 # import foyer
 import os
-import pathlib
+from pathlib import Path
 
 import flow
-from flow import environments
+from flow import directives
+
+import reproducibility_project.templates.ndcrc
```
justinGilmer, line 9: This part is a bit out of my current domain knowledge. But do you need to do anything like instantiate the `reproducibility_project.templates.ndcrc.NotreDameCRC` object in your project.py? Or does this just work for you?

justinGilmer (commented):
Few small comments, and it LGTM!

reproducibility_project/templates/crc.nd.sh
rwsmith7531: I don't know all about how that works.  I only changed a few things from what Ryan DeFever had in another workflow.

reproducibility_project/src/engines/cassandra/project.py
rwsmith7531: Nope, it just works.  This is apparently because `ComputeEnvironment` is a meta-class, according to https://docs.signac.io/en/latest/environments.html#how-to-use-environments .

reproducibility_project/templates/crc.nd.sh
rwsmith7531: It seems to work as-is, and I'd rather not mess with it if it isn't broken, since I'm not familiar with how it works.

justinGilmer: Fair enough!

reproducibility_project/src/engines/cassandra/project.py
justinGilmer: That is awesome, and I learned something new!

justinGilmer (approved):

PR #72, opened 2021-08-23
Create plots with t0 as a vertical line

justinGilmer:
To make quick inspection of trajectories and their progress towards
equilibration, being able to quickly generate plots over a data space to
inspect with signac dashboard will make our lives much easier.

This is not feature complete, features like matplotlib plotting kwargs
and other options will be exposed to the user, which will be implemented
in the analysis module.

Simple example using a `pymbar.testsystems.correlated_timeseries_example(N=10000, tau=200)
test

The vertical line is the t0 that maximizes the amount of uncorrelated samples from `pymbar.detectEquilibration`

### DIFF
--- a/reproducibility_project/src/analysis/equlibration.py
+++ b/reproducibility_project/src/analysis/equlibration.py
@@ -1,9 +1,12 @@
 """Timeseries and pyMBAR related methods."""
+import pathlib
 from typing import List
 
+import matplotlib.pyplot as plt
 import numpy as np
 import numpy.typing as npt
 from pymbar import timeseries
+from unyt import matplotlib_support
 
 
 def is_equilibrated(
@@ -20,7 +23,7 @@ def is_equilibrated(
     'production' region from 'a_t'. The fraction of 'production' data is
     then compared to the threshold value. If the fraction of 'production' data
     is >= threshold fraction this will return a list of
-    [True, t0, g] and [False, None, None] otherwise.
+    [True, t0, g, Neff] and [False, None, None, None] otherwise.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


reproducibility_project/src/analysis/equlibration.py
```diff
@@ -76,10 +80,54 @@ def trim_non_equilibrated(
...
+
+    _, t0, g, Neff = is_equilibrated(a_t, threshold=threshold, nskip=1)
+
+    ymin = np.min(a_t) * scale
+    ymax = np.max(a_t) * scale
+
+    plt.plot(
+        a_t,
+    )
+    plt.vlines(x=t0, ymin=ymin, ymax=ymax)
```
chrisjonesBSU, line 130: It might be better to do a dashed line style and a color other than the blue matplot lib defaults to.

jennyfothergill: Agreed. I think we can hardcode the dash, but if we want the color to change with the rc file (#74) we can use [cycler](https://matplotlib.org/stable/tutorials/intermediate/color_cycle.html)

justinGilmer: I like this idea. Ill begin with a basic hardcoded one, but we can make this a bit more programmatic later.

jennyfothergill (approved):

justinGilmer:
test
Not the best, but for an initial and quick screen, should be good. Next steps will be in a separate PR which will operate on a job, and use statepoint info as additional formatting.

justinGilmer:
Ive cleaned things up a bit. @jennyfothergill or anyone else who already has a "log.txt" file, can one of yall test this to make sure it is behaving as expected?

justinGilmer:
Just tested on a sample log.txt from @jennyfothergill , seems to be working well!

jennyfothergill:
> test
> Not the best, but for an initial and quick screen, should be good. Next steps will be in a separate PR which will operate on a job, and use statepoint info as additional formatting.

Forgive me if this is a nitpick, but Neff is the number of uncorrelated samples, right? Let's format it as an integer.

justinGilmer:
Completely correct and I have changed this now. I also had some bugs where dictionaries were being passed by reference, and would be left with previous values in the label if you tried to plot multiple properties in a single session (like an ipython instance). This has been fixed. I also reduced the scaling and made it behave much better than before.

PR #77, opened 2021-08-23
Fix a typo in the equilibration filename

jennyfothergill:


justinGilmer (approved):
🤦‍♂️  Great catch, Jenny!

PR #79, opened 2021-08-23
Add diffusion coefficient calculation

jennyfothergill:
related to #4 

Freud's MSD class requires the particle image if any particle crosses the PBC.

### DIFF
--- a/reproducibility_project/src/analysis/diffusion.py
+++ b/reproducibility_project/src/analysis/diffusion.py
@@ -1,0 +1,61 @@
+"""Facilitate calculation of the diffusion coefficient from simulation data."""
+
+import freud
+import gsd.hoomd
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+def gsd_msd(job, skip=2, stride=1):
+    """Compute the MSD given a Signac Job object.
+
+    The job folder is expected to contain the file "trajectory.gsd" with lengths
+    in nanometers.
+    This function is a convenience wrapper for freud's MSD module
+    https://freud.readthedocs.io/en/latest/modules/msd.html
+    After execution, the files "msd.png" and "msd.txt" are created in the job
+    folder.
+
+    Parameters
+    ----------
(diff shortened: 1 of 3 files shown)


justinGilmer:
Not entirely sure if this is completely ok to do. But it seems like the [freud msd calculation](https://freud.readthedocs.io/en/latest/modules/msd.html#freud.msd.MSD) can either be provided a box that is assumed constant beforehand, or it can take already-unwrapped data and passing in `None` for box.

I wonder if that would work for this case?

reproducibility_project/src/analysis/diffusion.py
```diff
@@ -0,0 +1,63 @@
...
+    -------
+    freud.msd.MSD
+        Computed MSD object
+    """
+    msd = _gsd_msd(job.fn("trajectory.gsd"), skip, stride)
+
+    fig, ax = plt.subplots()
+    ax.plot(msd.msd)
+    ax.set_xlabel("$Frame$")
+    ax.set_ylabel(r"$MSD (nm^{2})$")  # TODO check unit
```
justinGilmer, line 39: Also, if we have `unyt` units associated with these values, we can use [unyt's built-in matplotlib support in a context manager format](https://unyt.readthedocs.io/en/stable/usage.html#plotting-with-matplotlib)

jennyfothergill:
> Not entirely sure if this is completely ok to do. But it seems like the freud msd calculation can either be provided a box that is assumed constant beforehand, or it can take already-unwrapped data and passing in None for box.

Yes, if folks want to unwrap their simulation themselves, that will also work fine. I will aim to design any function with that in mind.

reproducibility_project/src/analysis/diffusion.py
jennyfothergill: wow, that's cool!

jennyfothergill: OK, I did try this, but it got a little wonky with the linear fit.

Issue #75, opened 2021-08-23
Sampling rates


justinGilmer:
What sampling rates do we want to use for our properties and trajectories in our simulations?

### NOTE: Suggestions for MD
trajectory_sample_rate: 10000 timesteps
property_sample_rate: 1000 timesteps

Just a generic guess at first, but the sooner we decide on this the better. I think @bc118 mentioned that the MC folks have defined values for their sims already.

bc118:
The Mc sampling and running values are listed in the excel sheet

jennyfothergill:
That sounds good to me--right now hoomd is running with 10,000 (trajectory) and 5,000 (property) as the interval between samples, but it's easy to change. :)

jennyfothergill:
I've added these intervals to [the project board](https://github.com/mosdef-hub/reproducibility_study/projects/1#card-66677615)

PR #80, opened 2021-08-23
Update GROMACS workflow

daico007:
Update the sampling interval per #75. Add an extra step in `gromacs/project.py` to extend unequilibrated simulations. Also combine `grompp` and `mdrun` into one step.

### DIFF
--- a/reproducibility_project/src/engines/gromacs/project.py
+++ b/reproducibility_project/src/engines/gromacs/project.py
@@ -4,9 +4,11 @@ import pathlib
 import sys
 
 import flow
+import panedr
 import unyt as u
 from flow import environments
 
+from reproducibility_project.src.analysis.equilibration import is_equilibrated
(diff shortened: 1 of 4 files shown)


reproducibility_project/src/engines/gromacs/project.py
```diff
@@ -208,6 +195,13 @@ def _setup_mdp(fname, template, data, overwrite=False):
     return None
 
 
+def equil_status(job, op, att):
+    """Check equilibrication status of specific attributes of specific operation."""
```
justinGilmer, line 199:
```suggestion
    """Check equilibration status of specific attributes of specific operation."""
```

justinGilmer (approved):
LGTM, one small spelling change

reproducibility_project/src/engines/gromacs/project.py
daico007: Fixed

PR #81, opened 2021-08-24
Mc3s simulations

ramanishsingh:


reproducibility_project/src/engines/mcccs/dlt
```diff
@@ -0,0 +1,475 @@
+"""Setup for signac, signac-flow, signac-dashboard for this study."""
```
justinGilmer, line 1: Is this file supposed to be here?

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -72,33 +100,76 @@ def has_fort_files(job):
...
+            job.doc.files_ready = True
+
+        return job.doc.files_ready
 
 
 @Project.label
 def has_restart_file(job):
     """Check if the job has a restart file."""
-    return job.isfile("fort.77")
+    return job.sp.ensemble == "GEMC-NVT" or job.isfile("fort.77")
```
justinGilmer, line 172: This will evaluate to true in either of these situations, is that was is expected?

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -165,59 +246,130 @@ def cool_finished(job):
...
+    """Sanitize the output logs for NPT simulations."""
+    import numpy as np
+
+    files = glob("fort*12*{}*".format(step))
+    arrays = []
+    for filecurrent in files:
+        array = np.genfromtxt(filecurrent, skip_header=1)
+        arrays.append(array)
+    arrays = np.vstack(arrays)
+    arrays[:, 0] = arrays[:, 0] / 10
```
justinGilmer, line 276: Lets add a comment here mentioning what these units are being converted to, or use `unyt`

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -165,59 +246,130 @@ def cool_finished(job):
...
+    os.chdir(job.ws)
+    files = glob("fort*12*{}*".format("equil"))
+    if len(files) < 2:  # at least do two loops of equilibration
+        print("equils done is less than 2")
+        return False
+    if job.sp.ensemble == "NPT":
+        equil_log = sanitize_npt_log("equil")
+
+    # run pymbar
+    return True
```
justinGilmer, line 296: You dont need to use os.chdir since you have the job.

you can do it in multiple ways

Refer to: https://docs.signac.io/projects/flow/en/latest/api.html#flow.with_job

```suggestion

def system_equilibrated(job):
    """Check if the system is equilibrated."""
    with job:
        files = glob("fort*12*{}*".format("equil"))
        if len(files) < 2:  # at least do two loops of equilibration
            print("equils done is less than 2")
            return False
        if job.sp.ensemble == "NPT":
            equil_log = sanitize_npt_log("equil")

    # run pymbar
    return True
```

or

```suggestion
@flow.with_job
def system_equilibrated(job):
    """Check if the system is equilibrated."""
    files = glob("fort*12*{}*".format("equil"))
    if len(files) < 2:  # at least do two loops of equilibration
        print("equils done is less than 2")
        return False
    if job.sp.ensemble == "NPT":
        equil_log = sanitize_npt_log("equil")

    # run pymbar
    return True
```

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -369,21 +607,24 @@ def run_cool(job):
...
     """Run equilibration."""
     from subprocess import PIPE, Popen
 
-    step = "equil"
-    """Run the melting stage."""
+    step = "equil" + str(job.doc.equil_replicates_done)
+    """Run the  equil  stage."""
     print("Running {}".format(step))
     execommand = "/home/rs/group-code/MCCCS-MN-7-20/exe-8-20/src/topmon"
     os.chdir(job.ws)
```
justinGilmer, line 626: Same as above
Just decorate:

```py3
@Project.post(system_equilibrated)
@flow.with_job
def run_equil(job):
```

justinGilmer (changes requested):
Some changes and questions, but it is looking great!

reproducibility_project/src/engines/mcccs/project.py
ramanishsingh: Fixed.

reproducibility_project/src/engines/mcccs/project.py
ramanishsingh: Thanks.

reproducibility_project/src/engine_input/mcccs/fort77maker_twobox.py
```diff
@@ -0,0 +1,457 @@
...
+
+    Returns
+    -------
+    NA
+    """
+    if not (len(name_list) == coordinates.shape[0]):
+        print(
+            "Error in xyzwriter. The atom list size and coordinate size is not same"
+        )
+        return
```
jennyfothergill, line 55:
```suggestion
        raise IndexError("The atom and coordinate lengths are different.")
```
Might be better to raise an error here instead of printing and returning as normal. I'm not sure if index error is the best choice, but it is my first thought.

reproducibility_project/src/engine_input/mcccs/fort77maker_twobox.py
```diff
@@ -0,0 +1,457 @@
...
+
+    Parameters
+    ----------
+    name_list : List containing element names
+    coordinates : xyz coordinates of elements in AA
+    out_filename : Name of the output filename
+
+    Returns
+    -------
+    NA
```
jennyfothergill, line 49:
```suggestion
```
If nothing is returned, this section can be omitted. This applies to other instances in the script, but I will mention this one.

reproducibility_project/src/engine_input/mcccs/fort77maker_twobox.py
```diff
@@ -0,0 +1,457 @@
...
+        output_string += "{} {} {} {}\n".format(
+            name_list[i],
+            coordinates[i, 0],
+            coordinates[i, 1],
+            coordinates[i, 2],
+        )
+    with open(out_filename, "w") as text_file:
+        text_file.write(output_string)
+    print("File written as {}".format(out_filename))
+    return
```
jennyfothergill, line 69:
```suggestion
```
If nothing is returned, this is not needed. Same for other instances.

reproducibility_project/src/engine_input/mcccs/fort77maker_twobox.py
```diff
@@ -0,0 +1,457 @@
...
+    fileBox[1] = xyz_file[0]
+    fileBox[2] = xyz_file[1]
+
+    #########config file names
+    config_file = {}
+    # INPUT ENDS HERE
+
+    # INPUT ENDS HERE
+
+    # INPUT ENDS HERE
```
jennyfothergill, line 197: Are we expecting the user to edit this script?

reproducibility_project/src/engine_input/mcccs/fort77maker_twobox.py
```diff
@@ -0,0 +1,457 @@
...
+    for x in list1:
+        # check if exists in unique_list or not
+        if x not in unique_list:
+            unique_list.append(x)
+    # print list
+    return unique_list
+
+
+# This function takes name_list, coordinates, and writes the xyz file in out_filename
+def xyzwriter(name_list, coordinates, out_filename):
```
jennyfothergill, line 38: Why not use `mbuild.formats.xyz.write_xyz`?

reproducibility_project/src/engines/mcccs/fort77maker_twobox.py
```diff
@@ -0,0 +1,457 @@
...
+from collections import Counter
+from subprocess import call
+
+import mbuild as mb
+import numpy as np
+
+print(sys.executable)
+
+
+def unique(list1):
```
jennyfothergill, line 14: is this duplicate?

reproducibility_project/src/engines/mcccs/fort77maker_twobox.py
```diff
@@ -0,0 +1,457 @@
...
+    for x in list1:
+        # check if exists in unique_list or not
+        if x not in unique_list:
+            unique_list.append(x)
+    # print list
+    return unique_list
+
+
+# This function takes name_list, coordinates, and writes the xyz file in out_filename
+def xyzwriter(name_list, coordinates, out_filename):
```
jennyfothergill, line 38: is this duplicate?

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -23,40 +23,77 @@ def __init__(self):
...
-    import warnings
+    import sys
 
-    import mbuild as mb
+    import foyer
+    from constrainmol import ConstrainedMolecule
 
-    from reproducibility_project.src.utils.forcefields import load_ff
+    sys.path.append(Project().root_directory())
+    from src.utils.forcefields import load_ff
```
jennyfothergill, line 39: check if this is still necessary after the pip install update.

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -23,40 +23,77 @@ def __init__(self):
...
-    warnings.simplefilter("ignore")
-    methane = mb.Compound(name="MET")
-    methane_child_bead = mb.Compound(name="_CH4")
-    methane.add(methane_child_bead, inherit_periodicity=False)
-    box_liq = mb.fill_box(
-        compound=methane, n_compounds=1230, box=[4.5, 4.5, 4.5]
-    )
+    sys.path.append(Project().root_directory() + "/src/molecules/")
+    sys.path.append(Project().root_directory() + "/..")
+    from system_builder import construct_system
```
jennyfothergill, line 43: again, this should be fixed with the pip install update

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -23,40 +23,77 @@ def __init__(self):
...
+
+    system = construct_system(job.sp)
+    molecule = get_molecules(job)[0]
```
jennyfothergill, line 46: I am trying to understand what your workflow is here: it looks like sometimes in `construct_system` you want to build a system with the constrained molecules--is that correct? If that's the case, let's add this functionality to the `construct_system` function just in case other folks who may want a constrained molecule system can also use it. :)

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -109,14 +187,27 @@ def has_topmon(job):
...
     )
 
 
+@Project.label
+def equil_replicate_set(job):
+    """Check if number of equil replicates done has been set."""
+    try:
+        return isinstance(job.doc.equil_replicates_done, int)
+    except AttributeError:
+        return False
```
jennyfothergill, line 208:
```suggestion
    return isinstance(job.doc.get("equil_replicates_done"), int)
```
I think you can simplify this logic like this. The `get` function will return `None` if the key does not exist, which will return `False` from `isinstance(None, int)`

jennyfothergill (changes requested):
I am only reviewing the python files, as I am not familiar enough with MCCCS or Fortran to review those files. 😅 
I've made a couple of style suggestions/simplifications, but my main suggestion is to modify the construct_system function to add constrainmol.

reproducibility_project/src/engine_input/mcccs/fort77maker_twobox.py
ramanishsingh: I made my own xyz write because mBuild only gives 6 decimal digits. I need at least 7 for my software because when the bonds are rigid, all bonds must be of the exact same length, with a precision of 1e-7 \AA.

reproducibility_project/src/engine_input/mcccs/fort77maker_twobox.py
ramanishsingh: Thanks! Should be removed. 

I made this script (restart file write for MCCCS) so that other people in the group can also use it for their simulations. And then pasted it here. Will delete these unnecessary lines.

reproducibility_project/src/engines/mcccs/project.py
ramanishsingh: Fixed. Thanks!

reproducibility_project/src/engines/mcccs/fort77maker_twobox.py
ramanishsingh: Yes. Fixed.

reproducibility_project/src/engines/mcccs/project.py
ramanishsingh: Works without adding the path.Thanks

reproducibility_project/src/engine_input/mcccs/fort77maker_twobox.py
ramanishsingh: Added.

reproducibility_project/src/engine_input/mcccs/fort77maker_twobox.py
jennyfothergill: hmm. Maybe this should be a kwarg to our xyz writer.

PR #93, opened 2021-08-31
Feat/neff as equilibration option

justinGilmer:
Previously, the only metric in `is_equilibrated` to determine if a set
of data was sufficiently equilibrated was only with a fraction of the
total dataset.

This creates issues when dealing with very small data sets since we also
require a sufficient amount of effectively uncorrelated samples to do
proper statistical treatments. Now, there are two values that need to be
true before we classify a dataset as equilibrated:

1. A fraction of the dataset must be considered effectively decorrelated
(the starting index of the location in the data that is the first
uncorrelated sample)
2. A minimum amount of uncorrelated samples needs to be available based
   on the starting index and the statistical inefficiency of the data.

This PR adds number 2, since number 1 was included in the first
iteration of this method (it has been named from `threshold` to
`threshold_fraction`.
A new function argument, `threshold_neff` has been added, with a default
value of 100.

Now, for `is_equilibrated` to return `True`, it must satisfy both of
these threshold values to be considered equilibrated.

### DIFF
--- a/reproducibility_project/src/analysis/equilibration.py
+++ b/reproducibility_project/src/analysis/equilibration.py
@@ -10,7 +10,11 @@ from signac.contrib.job import Job
 
 
 def is_equilibrated(
-        a_t: npt.ArrayLike, threshold_fraction: float = 0.8, threshold_neff: int = 100, nskip: int = 1,) -> List:
+    a_t: npt.ArrayLike,
+    threshold_fraction: float = 0.8,
+    threshold_neff: int = 100,
+    nskip: int = 1,
+) -> List:
     """Check if a dataset is equilibrated based on a fraction of equil data.
 
     Using `pymbar.timeseries` module, check if a timeseries dataset has enough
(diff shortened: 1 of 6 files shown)


reproducibility_project/src/analysis/equilibration.py
```diff
@@ -10,14 +10,17 @@
...
+    a_t: npt.ArrayLike,
+    threshold_fraction: float = 0.8,
+    threshold_neff: int = 100,
+    nskip: int = 1,
 ) -> List:
     """Check if a dataset is equilibrated based on a fraction of equil data.
 
     Using `pymbar.timeseries` module, check if a timeseries dataset has enough
-    equilibrated data based on a threshold value. The threshold value
+    equilibrated data based on one or two threshold values. The threshold_fraction value
```
justinGilmer, line 21:
```suggestion
    equilibrated data based on two threshold values. The threshold_fraction value
```

daico007 (approved):
LGTM!

PR #94, opened 2021-08-31
Add NPT->NVT workflow for HOOMD

jennyfothergill:
- [x] check that NPT is equilibrated before starting NVT
- [x] add a shrink step to NVT to adjust to the average volume found by NPT
- [x] add restart logic if NPT/NVT is not equilibrated


reproducibility_project/src/engines/hoomd/project.py
```diff
@@ -36,29 +34,69 @@ def add_args(cls, parser):
         parser.add_argument("--nodelist", help="Specify the node to submit to.")
 
 
-# This environment variable is set by running
+# The MOSDEF_PYTHON environment variable is set by running
 # echo "export MOSDEF_PYTHON=$(which python)" >> ~/.bashrc
 # with the mosdef-study38 conda env active
 @Project.operation.with_directives({"executable": "$MOSDEF_PYTHON", "ngpu": 1})
```
justinGilmer, line 40: just so i understand, what is the importance of this section with the env variable?

reproducibility_project/src/engines/hoomd/project.py
```diff
@@ -104,45 +170,90 @@ def run_hoomd(job):
...
+
+    data = np.genfromtxt(job.fn(f"log-{method}.txt"), names=True)
+    prop_data = data[eq_property]
+    iseq, _, _, _ = eq.is_equilibrated(prop_data)
+    if iseq:
+        uncorr, i, g, N = eq.trim_non_equilibrated(prop_data)
+        job.doc[f"avg_{eq_property}"] = np.average(uncorr)
+        job.doc[f"std_{eq_property}"] = np.std(uncorr)
+    job.doc[f"{method}_eq"] = iseq
+    return iseq
```
justinGilmer, line 256: I dont think this is doing what we expect.

`trim_non_equilibrated` returns the dataset `a_t` that begins from `t0`, so you still need to run `subsampleCorrelatedData` on this. At least for the avg properties? Or maybe this is just a quick check?

justinGilmer (commented):
Two quick questions and then this should be good to go!

reproducibility_project/src/engines/hoomd/project.py
jennyfothergill: in order to use a conda environment on our cluster, we have to provide the path to that python executable--for me this would be something like `/home/jennyfothergill/... ` our conda environment python executables are in our user directory. If I set this code to run on my specific conda installation and @chrisjonesBSU wanted to also run this project, he would have to specify a different path. By using this environment variable which is set in a standard way, we can each set it separately.

reproducibility_project/src/engines/hoomd/project.py
jennyfothergill: oops! you are correct. I will fix this now

tcmoore3 (commented):
LGTM

justinGilmer (approved):

PR #95, opened 2021-09-01
Write out subsampled data points to job.data

justinGilmer:
We currently have the ability to write out the indices of the subsampled
data that is effectively decorrelated.

This stores the values of the subsampled property in the `job.data`
object (see https://docs.signac.io/en/latest/jobs.html#job-data-storage
) for more information.

The values of the property are located at:
`job.data["subsamples/PROPERTY_OF_INTEREST"]`. For example:
`job.data["subsamples/density"]`. This provides a numpy-like access to
tabular data and scalar values.

This is currently missing unit tests.


reproducibility_project/src/analysis/sampler.py
```diff
@@ -42,6 +44,46 @@ def sample_job(
...
+
+    if (
+        not overwrite
+        and job.data.get(f"subsamples/{property}", None) is not None
+    ):
+        raise ValueError(
+            f"Attempting to overwrite already existing data for property: {property}, set overwrite=True to do this."
+        )
+
+    sampling_indices = job.doc["sampling_results"][f"{property}"]
```
justinGilmer, line 72: Note: We cannot serialize a python `range` object to json. So we need to also update `write_subsampled_values` above

```suggestion
    sampler_output= job.doc["sampling_results"][f"{property}"]
```

reproducibility_project/src/analysis/sampler.py
```diff
@@ -42,6 +44,46 @@ def sample_job(
     job.doc["sampling_results"][variable] = (range(start, stop, step), Neff)
```
justinGilmer, line 44:
```suggestion
    job.doc["sampling_results"][variable] = {"start": start, "stop": stop, "step": step, "Neff": Neff}
```

PR #100, opened 2021-09-01
Create xyz_rdf analysis function for Cassandra trajectories.

rwsmith7531:


jennyfothergill:
@rwsmith7531 Does the xyz file provided have unwrapped coordinates?

rwsmith7531:
> @rwsmith7531 Does the xyz file provided have unwrapped coordinates?

No, does it need unwrapped?  It's wrapped with molecules intact.

jennyfothergill:
@rwsmith7531 sorry, no I misspoke--but you will need the box information. Does the MDAnalysis xyz reader preserve the box information?

rwsmith7531:
@jennyfothergill no, it doesn't, because there's no box information to preserve in xyz files.  Cassandra does output box info, but in a different file.  I can probably make the function read the box info from there and store it in the MDAnalysis timesteps.  Thanks for catching that.

jennyfothergill:
It might even be worthwhile, since you'll have to do some converting anyway, to convert your output to gsd format. Then you can simply use the same analysis functions.  Check out [this card](https://github.com/mosdef-hub/reproducibility_study/projects/1#card-66677615) and [this issue](https://github.com/mosdef-hub/reproducibility_study/issues/25) for more info. I'm happy to help with this if you run into any issues.

rwsmith7531:
@jennyfothergill I was going to add an auxiliary reader for `.H` files to use to create a MDAnalysis xyz trajectory reader that includes box info from the .H file.  I looked at what mdtraj has, and it doesn't seem to have many tools for gsd files, it doesn't read .H files, and it doesn't appear to convert xyz to gsd without loading the whole trajectory into memory at once.  I see you offered to help @ramanishsingh with this; did you come up with something useful for converting xyz to gsd?  Am I missing something?

jennyfothergill:
[Here](https://gist.github.com/jennyfothergill/a8163b0e4445fc1103ff7167bb625c5d) is a snippet for converting Cassandra output to gsd

Issue #97, opened 2021-09-01
Trajectory Name For Diffusion Calculations


CalCraven:
Looking at the calculation for RDF and for diffusion, both apply their analysis for default trajectory names of `trajectory.gsd`. However, the diffusion calculation needs to be applied to an NVT production simulation, whereas the RDF does not have that stipulation. Unless the plan is to also do all of our analysis from the NVT simulation, instead of the "production" simulation that is in NPT, we might want to allow for a keyword to be passed to these [analysis routines](https://github.com/mosdef-hub/reproducibility_study/blob/main/reproducibility_project/src/analysis/rdf.py) to make sure the correct trajectory is grabbed. It also might be nice to add to the project some guidelines for naming conventions so we all keep relatively similar.

Here are links to the relevant lines of code:
https://github.com/mosdef-hub/reproducibility_study/blob/8b24e488fb0e7e431792c939ed4c145004d80d0a/reproducibility_project/src/analysis/rdf.py#L42
https://github.com/mosdef-hub/reproducibility_study/blob/8b24e488fb0e7e431792c939ed4c145004d80d0a/reproducibility_project/src/analysis/diffusion.py#L43

jennyfothergill:
I agree--when I wrote these, I had less of a grasp on the whole project and I used this as a placeholder. 
In hoomd I am using the names "trajectory-npt.gsd", "trajectory-nvt.gsd", and "log-npt.txt", "log-nvt.txt".

related to #89

jennyfothergill:
fixed by #101

PR #101, opened 2021-09-02
Add filenames as kwargs to analysis functions

### LINKED ISSUE #89: Option to change log filename in `plot_job_property_with_t0`
Currently the function only reads file named "log.txt".
This might not work for systems with two boxes, where we have something like "log_box1.txt", and "log_box2.txt".
New optional argument, `log_filename` can be introduced in the function.

jennyfothergill:
fixes #89 #97 

We should still use some [standard file conventions](https://github.com/mosdef-hub/reproducibility_study/projects/1#card-67959147) as @CalCraven suggested, but this PR allows for more flexibility when using the analysis functions.

### DIFF
--- a/reproducibility_project/src/analysis/diffusion.py
+++ b/reproducibility_project/src/analysis/diffusion.py
@@ -9,7 +9,7 @@ from freud.msd import MSD
 from scipy import stats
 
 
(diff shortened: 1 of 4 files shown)


CalCraven (approved):
LGTM, should be good to implement.

justinGilmer (approved):

PR #105, opened 2021-09-10
Include analysis routines to write sampled data

justinGilmer:
Currently, we have utilities to subsample the properties of interest for
statistical treatment, however, we have not implemented this in our
analysis signac project.

This is the initial addition of these routines for both NPT and NVT
production data.

Additional unit tests are needed, as well as further normalization of
the data between each engine (specifically column names), but since we
have the data for methane manually calculated, this should be a great
way to test this.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
   rev: 21.8b0
   hooks:
     - id: black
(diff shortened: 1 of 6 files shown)


zijiewu3:
Otherwise LGTM. Another suggestion might be to add several labels checking that the property values have been written to the `doc`.

PR #106, opened 2021-09-10
Add cassandra2gsd function.

rwsmith7531:
This PR adds a function to convert a Cassandra trajectory to gsd format.

### DIFF
--- a/reproducibility_project/src/utils/trajectory_conversion.py
+++ b/reproducibility_project/src/utils/trajectory_conversion.py
@@ -39,7 +39,7 @@ def cassandra2gsd(Hpath, xyzpath, gsdpath, species_list):
                     buff.write(xyzfile.readline())
                 buff.seek(0)
                 xyz = np.loadtxt(buff, usecols=(1, 2, 3)) * 0.1
-            if all(nmols != nmols_old):
+            if any(nmols != nmols_old):
                 typed_system = parmed.Structure()
                 for i, parspec in enumerate(species_list):
                     n = nmols[i]
--- a/reproducibility_project/src/utils/trajectory_conversion.py
+++ b/reproducibility_project/src/utils/trajectory_conversion.py
@@ -7,9 +7,17 @@ import hoomd
 import mbuild as mb
 import numpy as np
 import parmed
+import unyt as u
 
 
-def cassandra2gsd(h_path, xyz_path, gsd_path, species_list):
+def cassandra2gsd(
+    h_path,
+    xyz_path,
+    gsd_path,
+    species_list,
+    gsd_length_unit="nm",
+    cassandra_length_unit="angstrom",
+):
     """Convert Cassandra H and xyz files to a gsd trajectory file.
 
     Inputs:
@@ -23,8 +31,24 @@ def cassandra2gsd(h_path, xyz_path, gsd_path, species_list):
             with one element per species.  This should be the same as the species_list
             supplied to the MoSDeF Cassandra System and MoveSet objects used to generate
             the trajectory.
-
+        gsd_length_unit: string or unyt.Unit representing the unit of length to be
+            used in the .gsd file.
+        cassandra_length_unit: string or unyt.Unit representing the unit of length
+            used in the .H and .xyz files.
     """
+    gsd_length_unit = u.Unit(gsd_length_unit)
+    if not (gsd_length_unit.dimensions is u.dimensions.length):
+        raise ValueError(
+            "gsd_length_unit must be a unit of length, not "
+            + str(gsd_length_unit)
+        )
+    cassandra_length_unit = u.Unit(cassandra_length_unit)
+    if not (cassandra_length_unit.dimensions is u.dimensions.length):
+        raise ValueError(
+            "cassandra_length_unit must be a unit of length, not "
+            + str(cassandra_length_unit)
+        )
+    length_factor = (1.0 * cassandra_length_unit).to_value(gsd_length_unit)
     h_path = Path(h_path)
     xyz_path = Path(xyz_path)
     gsd_path = Path(gsd_path)
@@ -40,7 +64,7 @@ def cassandra2gsd(h_path, xyz_path, gsd_path, species_list):
                 for i in range(3):
                     buff.write(h_file.readline())
                 buff.seek(0)
-                lmat = np.loadtxt(buff) * 0.1
+                lmat = np.loadtxt(buff) * length_factor
             h_file.readline()
             nspecies_in_box = int(h_file.readline().strip())
             nmols = np.zeros(nspecies, dtype=int)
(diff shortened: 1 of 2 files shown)


reproducibility_project/src/utils/trajectory_conversion.py
```diff
@@ -0,0 +1,74 @@
...
+    """Convert Cassandra H and xyz files to a gsd trajectory file."""
+    Hpath = Path(Hpath)
+    xyzpath = Path(xyzpath)
+    gsdpath = Path(gsdpath)
+
+    nspecies = len(species_list)
+    nmols_old = np.zeros(nspecies, dtype=int)
+
+    with Hpath.open() as Hfile, xyzpath.open() as xyzfile, gsd.hoomd.open(
+        str(gsdpath), "wb"
```
jennyfothergill, line 21:
```suggestion
        gsdpath, "wb"
```
it is fine to pass a path object here as well

reproducibility_project/src/utils/trajectory_conversion.py
```diff
@@ -0,0 +1,74 @@
...
+
+import gsd.hoomd
+import mbuild as mb
+import numpy as np
+import parmed
+
+
+def cassandra2gsd(Hpath, xyzpath, gsdpath, species_list):
+    """Convert Cassandra H and xyz files to a gsd trajectory file."""
+    Hpath = Path(Hpath)
```
jennyfothergill, line 13: The convention in python is usually to use snake_case for class instances.
```suggestion
    h_path = Path(h_path)
```
it's fine to ignore this suggestion if you have strong feelings about it.

reproducibility_project/src/utils/trajectory_conversion.py
```diff
@@ -0,0 +1,74 @@
...
+
+    with Hpath.open() as Hfile, xyzpath.open() as xyzfile, gsd.hoomd.open(
+        str(gsdpath), "wb"
+    ) as gsdfile:
+        while Hfile.readline():
+            with io.StringIO() as buff:
+                for i in range(3):
+                    buff.write(Hfile.readline())
+                buff.seek(0)
+                lmat = np.loadtxt(buff) * 0.1
```
jennyfothergill, line 28: it might be nice to have this scaling factor as a kwarg

reproducibility_project/src/utils/trajectory_conversion.py
```diff
@@ -0,0 +1,74 @@
...
+import io
+from pathlib import Path
+
+import gsd.hoomd
+import mbuild as mb
+import numpy as np
+import parmed
+
+
+def cassandra2gsd(Hpath, xyzpath, gsdpath, species_list):
```
jennyfothergill, line 11: A little bit of info in the docstring about these parameters would be helpful, particularly species_list--I assume it is a list of pmd.Structures.

reproducibility_project/src/utils/trajectory_conversion.py
```diff
@@ -0,0 +1,74 @@
...
+                typed_system = parmed.Structure()
+                for i, parspec in enumerate(species_list):
+                    n = nmols[i]
+                    if n > 0:
+                        typed_system += parspec * n
+            bonds = [
+                (bond.atom1.idx, bond.atom2.idx) for bond in typed_system.bonds
+            ]
+            all_types = [a.type for a in typed_system.atoms]
+            types = list(set(types))
```
jennyfothergill, line 52:
```suggestion
            types = list(set(all_types))
```
This fails without this change

reproducibility_project/src/utils/trajectory_conversion.py
```diff
@@ -0,0 +1,74 @@
...
+    Hpath = Path(Hpath)
+    xyzpath = Path(xyzpath)
+    gsdpath = Path(gsdpath)
+
+    nspecies = len(species_list)
+    nmols_old = np.zeros(nspecies, dtype=int)
+
+    with Hpath.open() as Hfile, xyzpath.open() as xyzfile, gsd.hoomd.open(
+        str(gsdpath), "wb"
+    ) as gsdfile:
```
jennyfothergill, line 22: I'm not sure why these are being converted to Path objects before passing to the context manager.
```
with open(xyzpath) as xyzfile, ...
``` 
works just as well without it

jennyfothergill (changes requested):

reproducibility_project/src/utils/trajectory_conversion.py
rwsmith7531: Because that way it will take Path objects or strings.

reproducibility_project/src/utils/trajectory_conversion.py
rwsmith7531: I agree, I just haven't added that part yet.

reproducibility_project/src/utils/trajectory_conversion.py
rwsmith7531: Maybe, but that's just there to convert from angstroms to nanometers.

reproducibility_project/src/utils/trajectory_conversion.py
rwsmith7531: I fixed it, I think.

reproducibility_project/src/utils/trajectory_conversion.py
rwsmith7531: I've added it now.

rwsmith7531:
I don't know why the pre-commit CI keeps failing but not committing its changes.  It doesn't seem to be a docstring issue.  I have pre-commit installed locally, and it passes the local pre-commit checks.

reproducibility_project/src/utils/trajectory_conversion.py
rwsmith7531: It takes any path-like object this way.

jennyfothergill (approved):
LGTM

reproducibility_project/src/utils/trajectory_conversion.py
rwsmith7531: I added the option to choose the units.

ramanishsingh:
Sorry I am a little late to review this. 

Looks good to me. 
I have a question for @rwsmith7531 and @jennyfothergill 
I see that you are creating the bonds list. Would it be (or have been) easier if you just save your top file (.mol2) at the start of the simulation and use mdtraj to load xyz with mol2 file as the top file and save the loaded mdtraj traj as gsd?
It won't work for GEMC (because of changing number of molecules) but should work fine for _NpT_ (the case for which we are calculating _g(r)_)

ramanishsingh (approved):

rwsmith7531:
@ramanishsingh No, that would be harder.  Cassandra doesn't do anything involving mol2 files, and I don't think mol2 files include box dimensions.  MoSDeF Cassandra requires a list of parameterized single-molecule parmed structures anyway.  I also don't think mdtraj has a function that converts a mdtraj trajectory to gsd (it has a function that takes ndarrays, and the xyz file data may be too much to load into memory all at once).  We may not be computing the RDF for GEMC, but maybe we need GEMC trajectories for other analysis?

rwsmith7531:
> I don't know why the pre-commit CI keeps failing but not committing its changes. It doesn't seem to be a docstring issue. I have pre-commit installed locally, and it passes the local pre-commit checks.

@justinGilmer do you know what's going on with this?

justinGilmer:
pre-commit.ci run

justinGilmer:
Ah it looks like the mcccs project file is the cause of the issue. Ill make a new PR that fixes that and then updating from upstream main should fix this @rwsmith7531

ramanishsingh:
Sorry! :D

justinGilmer:
pre-commit.ci run

justinGilmer:
pre-commit.ci autofix

PR #109, opened 2021-09-10
Set tail correction to F in MCCCS simulations

ramanishsingh:


reproducibility_project/src/engines/mcccs/analysis.py
```diff
@@ -0,0 +1,89 @@
...
+def avg_one_seed_density(prod_run_files):
+    """For a one particular seed, read all the prod run files and provide the average density (g/ml) for one seed."""
+    if len(prod_run_files) == 0:
+        return None
+    os.system(
+        "grep 'specific density                        ' run.prod* | awk '{print $5'} > temp_density.txt"
+    )
+    density_list_one_seed = np.genfromtxt("temp_density.txt")
+    # now delete the temp file
+    os.remove("temp_density.txt")
```
justinGilmer, line 84: Since you are just using a temporary file that gets deleted once you make your numpy array. It is probably a bit better to use a [try except](https://docs.python.org/3/tutorial/errors.html) statement to ensure that you delete the file in the case of an error.
```suggestion
    os.system(
        "grep 'specific density                        ' run.prod* | awk '{print $5'} > temp_density.txt"
    )
    try:
        density_list_one_seed = np.genfromtxt("temp_density.txt")
    except Exception as e:
        raise e
     finally:
        # now delete the temp file
        os.remove("temp_density.txt")

```

Need to verify that my tabs are properly set, i would manually add this instead of accepting this commit.

justinGilmer (commented):
One small change and then this is g2g.

reproducibility_project/src/engines/mcccs/analysis.py
ramanishsingh: I have replaced it with

```python
    os.system(
        "grep 'specific density                        ' run.prod* | awk '{print $6'} > temp_density.txt"
    )
    if os.path.exists("temp_density.txt"):
        try:
            density_list_one_seed = np.genfromtxt("temp_density.txt")
        except Exception as e:
            raise e
        finally:
            # now delete the temp file
            os.remove("temp_density.txt")
        return np.mean(density_list_one_seed)
    else:
        return None
```

ramanishsingh:
Something terrible happened. I got so many junk files.

PR #110, opened 2021-09-13
Create subproject to explore LRC and shifting

justinGilmer:
Our initial results from the TraPPE United-Atom methane simulations
(`MethaneUA`) had discrepancies on the order of a few percentage points
for the density values when comparing MD vs MC simulations.

This appears to be a factor due to long range corrections (tail
corrections) and how the potential is truncated at the cutoff (abruptly
change to 0 once you reach r_cut) or shifted (a constant term is applied
to make the value of the potential at r_cut 0).

A subproject has been created to further explore this for MethaneUA for
all engines that support LRC and shifted potentials. This PR sets up the
groundwork for this.

### DIFF
--- a/reproducibility_project/lrc_shift_subproject/signac.rc
+++ b/reproducibility_project/lrc_shift_subproject/signac.rc
@@ -1,0 +1,3 @@
+project = lrc_shift_subproject
+workspace_dir = workspace
+schema_version = 1
--- a/reproducibility_project/lrc_shift_subproject/subproject-init.py
+++ b/reproducibility_project/lrc_shift_subproject/subproject-init.py
@@ -1,0 +1,233 @@
+"""Initialize signac statepoints."""
+import itertools
+import os
+
+import numpy as np
+import signac
+import unyt as u
+from numpy import ModuleDeprecationWarning
+
+molecules = ["methaneUA"]
+replicas = range(16)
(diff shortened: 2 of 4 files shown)


jennyfothergill (commented):
out of curiosity, why aren't we adding the tail correction information to the main workspace?

justinGilmer:
> out of curiosity, why aren't we adding the tail correction information to the main workspace?


At the moment, this is a subproject to study the affects of the tail correction and the cutoff shift or truncation and their affects on the results of our simulation.

We definitely might switch to those as the defaults, but this is just the exploratory study.

> EDIT: never mind--I think I understand. Are we to use the project.py files from the main workspace on the subproject?

And that can definitely work, although it is also ok to copy over the project.py files if you need to do any additional special checks/etc. Our `project-analysis.py` file should work without issue. We just have to be in this folder when running the analysis file.

jennyfothergill:
I was able to run this with the existing project.py file (although I needed to copy over the `templates` directory into the lrc folder) and have updated the google doc. I have the change necessary on the [feat/hoomd-lrc branch](https://github.com/jennyfothergill/reproducibility_study/tree/feat/hoomd-lrc) of my fork.

jennyfothergill (approved):

PR #113, opened 2021-09-15
Update Lammps project for nvt and npt simulations

CalCraven:
Include changes to nvt and npt simulations.
Run unit conversions to produce proper log and gsd files.
Potentially will also include the study information checking tail corrections and shifts for different molecules.


reproducibility_project/src/engines/lammps-VU/project.py
```diff
@@ -182,43 +196,128 @@ def lammps_em_nvt(job):
...
+        "kinetic_energy",
+        "pressure",
+        "temperature",
+        "density",
+    ]
+    # convert units
+    KCAL_TO_KJ = 4.184  # kcal to kj
+    ATM_TO_MPA = 0.101325  # atm to mpa
+    GPCM3_TO_AMUPNM3 = 0.6023  # g/cm^3 to amu/nm^3
+    df_in["pe"] = df_in["pe"] * KCAL_TO_KJ
```
justinGilmer, line 276: You define these multiple times below.

justinGilmer: Or, twice actually, not multiple

justinGilmer (changes requested):
Small change and then lgtm

justinGilmer:
Also need to run pre-commit locally or review the error message from pre-commit ci to fix the docstring.

justinGilmer (approved):

PR #115, opened 2021-09-16
Cassandra project update with collection and statistics scripts

emarinri:
Update the Cassandra project with computation of averages for NPT methane


justinGilmer (approved):

daico007 (approved):

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -109,26 +110,25 @@ def run_cassandra(job):
     cutoff = job.sp.r_cut * u.nm
 
     seedslist = [
-        [7860904, 98601355],
-        [955793508, 417823039],
-        [254420642, 2130720464],
-        [58120272, 1465850411],
-        [757616664, 1940492980],
```
rwsmith7531, line 116: Why did the seeds have to change?  I suppose it doesn't matter, as long as you use the same seedslist for each job (so that it's reproducible).  Did you use the new seedslist for all jobs?

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -376,6 +376,71 @@ def run_cassandra(job):
...
+        "enthalpy",
+    ]
+
+    NA = 6.032e23
+
+    C = ele.element_from_symbol("C")
+    H = ele.element_from_symbol("H")
+
+    mw = C.mass + H.mass * 4
+    mw *= g / mole
```
rwsmith7531, line 413: Why do you need the MW?  Cassandra already knows the molecular weight (it's also included in the parmed structures given to MoSDeF Cassandra) and outputs the mass density, so you can just use the output mass density in the prp file instead of converting from molecular density.  The way this is written, it would only work for methane.  See my suggested change later in the file.

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -376,6 +376,71 @@ def run_cassandra(job):
...
+        box1 = "prod.out.prp"
+
+        data_box1 = pd.read_table(
+            job.fn(box1), skiprows=3, sep="\s+", names=proplist
+        )
+
+        job.document.mean_energy_box1 = data_box1["energy_total"].mean()
+        job.document.mean_density_box1 = (
+            data_box1["density"].mean() / NA * mole / (angstrom ** 3)
+        ).to("mole/cm**3") * mw
```
rwsmith7531, line 441: You can convert this to a `unyt_quantity` if you really want to put a `unyt_quantity `in the job document, but you have to change your `unyt` import.  I just did it without unyt so that this suggested change can work on its own.  You can also remove `NA` and the MW stuff from the function.
```suggestion
        job.document.mean_density_box1 = 
            data_box1["mass_density"].mean() * 0.001
```

rwsmith7531 (changes requested):
I have some changes to suggest.

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -376,6 +376,71 @@ def run_cassandra(job):
...
+        "energy_inter",
+        "energy_intervdw",
+        "energy_lrc",
+        "energy_interq",
+        "energy_recip",
+        "energy_self",
+        "enthalpy",
+    ]
+
+    NA = 6.032e23
```
rwsmith7531, line 407: Avogadro's number is implemented incorrectly here.  I recommend removing its use entirely, but this is what it should be if you insist on using Avogadro's number.  When I correct for the wrong value, Cassandra's results are much closer to those of the other engines.
```suggestion
    NA = 6.032e23
```
```suggestion
    NA = 6.022e23
```

reproducibility_project/src/engines/cassandra/project.py
emarinri: I was getting an error from MoSDeF Cassandra regarding the length of the seeds [here](https://github.com/MaginnGroup/mosdef_cassandra/blob/cc0d3605a3b6f7a0c952324fb92a5aeb91032dfd/mosdef_cassandra/writers/inp_functions.py#L778).

reproducibility_project/src/engines/cassandra/project.py
emarinri: Sounds good!

rwsmith7531:
pre-commit.ci run

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -376,6 +376,69 @@ def run_cassandra(job):
...
+    else:
+
+        box1 = "prod.out.prp"
+
+        data_box1 = pd.read_table(
+            job.fn(box1), skiprows=3, sep="\s+", names=proplist
+        )
+
+        job.document.mean_energy_box1 = data_box1["energy_total"].mean()
+        job.document.mean_density_box1 = data_box1["mass_density"].mean() * 0.001
```
rwsmith7531, line 439:
```suggestion
        job.document.mean_energy_box1 = data_box1["energy_total"].mean()
        job.document.mean_density_box1 = (
            data_box1["mass_density"].mean() * 0.001
        )
```

reproducibility_project/src/engines/cassandra/project.py
rwsmith7531: Ah, that's probably due to the overflow that occurs in Cassandra if a seed goes over 2,147,483,648.  I made sure they didn't exceed that, but MoSDeF Cassandra won't let them exceed 1,000,000,000.

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -376,6 +376,72 @@ def run_cassandra(job):
...
+        "enthalpy",
+    ]
+
+    NA = 6.022e23
+
+    C = ele.element_from_symbol("C")
+    H = ele.element_from_symbol("H")
+
+    mw = C.mass + H.mass * 4
+    mw *= g / mole
```
rwsmith7531, line 413: This part isn't needed anymore, and would probably just serve to confuse other readers.
```suggestion

```

reproducibility_project/src/engines/cassandra/project.py
```diff
@@ -376,6 +376,72 @@ def run_cassandra(job):
         )
 
 
+@Project.operation
+@Project.pre(cassandra_complete)
+@Project.post(lambda job: "mean_energy_box1" in job.document)
+def statistics(job):
+    """Compute statistical quantities for each job."""
+    import ele
+    from unyt import angstrom, g, mole
```
rwsmith7531, line 385: You don't use unyt in this function, so you don't need to import from it.
```suggestion
```

rwsmith7531 (changes requested):
It looks good but I still request a couple changes for clarity of reading.

rwsmith7531 (approved):
LGTM.

PR #116, opened 2021-09-17
Remove statepoints with LRC and shifted potentials

justinGilmer:
It does not make sense to run a simulation with both long range
corrections and shifted potentials at the same time, in fact, [LAMMPS
forbids this entirely](https://docs.lammps.org/pair_modify.html#restrictions).

LRC's are meant to be used when there are discontinuities at the cutoff,
and shifted potentials are meant to transition the potential to 0 at the
cutoff. These attempt to mitigate similar issues.


CalCraven (approved):
LGTM!

PR #118, opened 2021-09-20
mc3s simulations project.py and analysis

ramanishsingh:
Creating a new PR because last one had a lot of junk files.
1. Set tail correction to F
2. Update project.py to use density for equilibration detection instead of box length
3. Include analysis.py file, that finds the mean and SEM from all replicas
4. Include conv_traj.py that combines the .xyz from all replicas into one gsd file for RDF calculation.


reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -909,6 +919,17 @@ def run_prod(job):
...
+            prod_file = text_file.write(
+                "All prod replicates done. Simulation finished for {} molecule = {}, ensemble = {}, temperature= {} K, pressure = {} kPa.".format(
+                    job,
+                    job.sp.molecule,
+                    job.sp.ensemble,
+                    job.sp.temperature,
+                    job.sp.pressure,
+                )
+            )
+            text_file.close()
```
justinGilmer, line 932: I recommend switching this to a `with open` statement just in case there is a failure while running, the file is guaranteed to be closed.

I would recommend not merging this suggestion though, and instead including it manually using your text editor, since I might have the tabs/spaces incorrect here.
```suggestion
            with open("production_information.txt", "w") as text_file:
                prod_file = text_file.write(
                    "All prod replicates done. Simulation finished for {} molecule = {}, ensemble = {}, temperature= {} K, pressure = {} kPa.".format(
                        job,
                        job.sp.molecule,
                        job.sp.ensemble,
                        job.sp.temperature,
                        job.sp.pressure,
                    )
                )
            text_file.close()
```

justinGilmer (changes requested):
One change and then LGTM

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -176,15 +186,19 @@ def equil_replicate_set(job):
...
 @Project.label
 def all_prod_replicates_done(job):
     """Check if all prod replicate simulations completed."""
     try:
-        a = job.doc.prod_replicates_done
-        b = job.doc.num_prod_replicates
+        a = job.doc.get("prod_replicates_done")
+        b = job.doc.get("num_prod_replicates")
+        if a >= b:
+            print("simulation complete for {} job".format(job))
```
justinGilmer, line 201: Im not sure if `NoneType` will work in a comparison like this. the `get` method will return `None` by default if one of those values are missing. If 0 is the lowest the number can be, I recommend:
```suggestion
        a = job.doc.get("prod_replicates_done", 0)
        b = job.doc.get("num_prod_replicates", 0)
        if a >= b:
            print("simulation complete for {} job".format(job))
```

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -316,166 +352,127 @@ def system_equilibrated(job):
...
-                and equil_status_energy2[0]
-                and equil_status_total_energy
-            ) == False:
-                print(
-                    "System {} is not equilibrated. Completed {} equil loops".format(
-                        job, job.doc.get("equil_replicates_done")
+                    text_file = open("equil_information.txt", "w")
+                    n = text_file.write(
+                        "Even though the system is not equilibrated according to pymbar, we are considering this system equilibrated as 3 equil loops are completed"
                     )
```
justinGilmer, line 454: I would convert this to a `with` statement too
```suggestion
                    with open("equil_information.txt", "w") as text_file:
                        n = text_file.write(
                            "Even though the system is not equilibrated according to pymbar, we are considering this system equilibrated as 3 equil loops are completed"
                    )
```

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -316,166 +352,127 @@ def system_equilibrated(job):
...
-                )
+                    text_file.close()
```
justinGilmer, line 455: If you use the `with` statement from above, then you can remove this line.

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -484,7 +481,13 @@ def system_equilibrated(job):
                 n = text_file.write(
                     "System {} is equilibrated at cycle {}. Completed {} equil loops".format(
                         job,
```
justinGilmer, line 483: Same here, this open can be a `with open` statement.

reproducibility_project/src/engines/mcccs/project.py
```diff
@@ -909,6 +919,17 @@ def run_prod(job):
...
+                prod_file = text_file.write(
+                    "All prod replicates done. Simulation finished for {} molecule = {}, ensemble = {}, temperature= {} K, pressure = {} kPa.".format(
+                        job,
+                        job.sp.molecule,
+                        job.sp.ensemble,
+                        job.sp.temperature,
+                        job.sp.pressure,
+                    )
+                )
+            text_file.close()
```
justinGilmer, line 932: The `with` statement will ensure that the file is closed automagically, no need to close a file manually when using them!
```suggestion
```

justinGilmer (changes requested):
Looking great, some small changes and then LGTM!

reproducibility_project/src/engines/mcccs/project.py
ramanishsingh: So in case the `prod_replicates_done` is not stored in job doc, will `job.doc.get("prod_replicates_done", 4)` return 4?

justinGilmer: Correct! The second value is the value that is returned if `get` cannot find that entry.

ramanishsingh: Thanks

justinGilmer (approved):

PR #119, opened 2021-09-20
Add rclone instructions

jennyfothergill:
Add instructions for rclone, linking to dropbox, cleaning the workspace.


README.md
```diff
@@ -28,3 +28,28 @@ MCCCS can be installed according to its documentation [here](https://ccs-psi.org
 
 ## Use
 This project uses the [Signac framework](https://signac.io/) to manage its parameter space. Instructions for initializing and submitting/running the simulation and analysis workflows can be found in [the project guide](reproducibility_project/README.md).
+
+## Sharing data
+In order to share data between contributors, we use [rclone](https://rclone.org/) to sync to a shared [dropbox](https://www.dropbox.com) folder.
+
+rclone can be installed without sudo permissions (as might be necessary, e.g., on a computing cluster) using conda.
```
jennyfothergill, line 35:
```suggestion
rclone can be installed without sudo permissions (as might be necessary, e.g., on a computing cluster) using conda:
```

reproducibility_project/clean_by_engine.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+echo Removing jobs from all engines but $1
```
CalCraven, line 2: Might be worth adding some check for a valid remove. Either that, or make someone specify exactly what they want to remove. 
i.e. you list the engines you want removed. For lammps-VU, I would remove [lammps-UD, gromacs, mcccs, gomc, hoomd, cassandra]. That way a typo doesn't accidentally delete all of the workspace.

reproducibility_project/clean_by_engine.sh
```diff
@@ -0,0 +1,10 @@
...
+for ENGINE in cassandra mcccs gomc gromacs hoomd lammps-VU lammps-UD
+do
+    if [[ $ENGINE != $1 ]]
+    then
+        signac rm $(signac find engine $ENGINE)
+        echo Removed jobs for $ENGINE
+    fi
+done
```
CalCraven, line 10: I would either add this to the lrc_cutoff_subproject, or allow an option to specify which subproject you want to filter other engine jobs from.

CalCraven (commented):
A few small quality of life changes, but otherwise it worked great for me!

reproducibility_project/clean_by_engine.sh
justinGilmer: That or you could make a `bin` folder at `reproducibility_study/reproducibility_project/bin` and put these scripts in here. Then the user can just run `../bin/clean_by_engine.sh my_engine` if they are in the sub project.

reproducibility_project/clean_by_engine.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+echo Removing jobs from all engines but $1
+for ENGINE in cassandra mcccs gomc gromacs hoomd lammps-VU lammps-UD
+do
+    if [[ $ENGINE != $1 ]]
+    then
+        signac rm $(signac find engine $ENGINE)
```
justinGilmer, line 7: Is there a verbose option for this? I think we would want to be verbose in this case so we see what is removed. Unless `signac rm` does this by default

justinGilmer (commented):
Two small suggestions and this LGTM!

reproducibility_project/clean_by_engine.sh
jennyfothergill: The verbose option for rm just echos the job id, which I don't find super helpful.

reproducibility_project/clean_by_engine.sh
jennyfothergill: yeah I like the idea of having a check with the user. I'll look into how to do this.

reproducibility_project/clean_by_engine.sh
justinGilmer: Makes sense to me. I think this is pretty good to go then.

reproducibility_project/clean_by_engine.sh
justinGilmer: I dont entirely know if this would work, but it might be one way to accomplish this
You could grep or use signac find on the workspace, if nothing shows up, then error?
https://unix.stackexchange.com/a/223504

jennyfothergill: @CalCraven I implemented your suggestion, so now it prompts the user before deleting anything, which feels much safer imo. :)

README.md
```diff
@@ -39,13 +39,13 @@ conda install -c conda-forge rclone
...
+Before syncing the workspace data, it is recommended to remove any empty job folders, as these will overwrite the data on dropbox. One way to do this is to remove the job folders of unused engines. We provide a bash script which can help(located in `reproducibility_study/reproducibility_project/bin/clean_by_engine.sh`). For example, to remove job folders for all engines EXCEPT hoomd, run:
 
 ```
-bash clean_by_engine.sh hoomd
+bash bin/clean_by_engine.sh hoomd
 ```
 
-Be careful when using this script as it could delete your data! Before using it is advisable to make a backup and make sure that you spell/capitalize your engine the same way as in the statepoint.
+Be careful when using this script as it could delete your data! Make sure so check which project (`mosdef_reproducibility` or `lrc_shift_subproject`) you are in using `signac project`. Before using it is advisable to make a backup and make sure that you spell/capitalize your engine the same way as in the statepoint.
```
justinGilmer, line 48:
```suggestion
Be careful when using this script as it could delete your data! Make sure to check which project (`mosdef_reproducibility` or `lrc_shift_subproject`) you are in using `signac project`. Before using it is advisable to make a backup and make sure that you spell/capitalize your engine the same way as in the statepoint.
```

reproducibility_project/clean_by_engine.sh
jennyfothergill: @justinGilmer The rm command will error if signac find returns nothing, so I don't think it's too dangerous. Putting rm in a script though definitely is a little scary though haha.

reproducibility_project/bin/clean_by_engine.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+echo Removing jobs from all engines but $1
+for ENGINE in cassandra mcccs gomc gromacs hoomd lammps-VU lammps-UD
+do
+    if [[ $ENGINE != $1 ]]
+    then
+        read -p "Delete $ENGINE job directories?[y/N] " yn
+        case $yn in
+            [Yy]* ) signac rm $(signac find engine $ENGINE); echo Removed jobs for $ENGINE;;
```
justinGilmer, line 9: What exactly does the `[Yy]*` mean here? the [Yy] im assuming is a regex pattern for lowercase or capital `Y`.

What does the * mean though?

I understand that the `* )` means the `else` case essentially.

jennyfothergill: > What exactly does the [Yy]* mean here?

Any input in [y, Y, yes, Yes, yep, Yankovic, yellow, etc] (anything starting with y or Y) will result in the directories being deleted.

> I understand that the * ) means the else case essentially.

Yep! Even if the user accidentally presses enter, nothing will be deleted.

justinGilmer: Perfect!

justinGilmer (approved):

PR #124, opened 2021-10-04
Update null value from None to "None" string

justinGilmer:
When trying to aggregate jobs according to various statepoint values, if
a `NoneType` value is used to represent that value not being necessary,
trying to aggregate jobs according to statepoints that have `NoneType`
as an option will fail due to the call of the `sort` method.

This makes it not possible to easily aggreate jobs when trying to do
final statistical analyses of our systems. This changes the default null
value to "None" string instead of `None`.


jennyfothergill (approved):
LGTM

PR #125, opened 2021-10-04
Cassandra project udpate

emarinri:
Updates for the current state of the LRC subproject for Cassandra

### DIFF
--- a/reproducibility_project/lrc_shift_subproject/src/engines/cassandra/project.py
+++ b/reproducibility_project/lrc_shift_subproject/src/engines/cassandra/project.py
@@ -1,0 +1,497 @@
+"""Setup for signac, signac-flow, signac-dashboard for this study."""
+# import foyer
+import os
+from pathlib import Path
+
+import flow
+import pandas as pd
+from flow import directives
+
+import reproducibility_project.templates.ndcrc
+
+
+class Project(flow.FlowProject):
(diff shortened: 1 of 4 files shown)


rwsmith7531 (approved):
LGTM.  Once this is merged with main, I intend to include it in the restart PR and make the same changes to the subproject `project.py`.

reproducibility_project/lrc_shift_subproject/src/engines/cassandra/project.py
```diff
@@ -0,0 +1,535 @@
...
+    cass_ensembles = {
+        "NPT": "npt",
+        "GEMC-NVT": "gemc",
+    }
+
+    cass_cutoffs = {
+        ("hard", None): "cut",
+        ("hard", "energy_pressure"): "cut_tail",
+        ("shift", None): "cut_shift",
+        ("shift", "energy_pressure"): None,
```
rwsmith7531, line 80: This isn't necessary since this will not be a statepoint, and if it were a statepoint, it would cause an error in Cassandra because `None` is not a valid cutoff style.

```suggestion
```

justinGilmer: @emarinri and @rwsmith7531 , once this suggestion is committed, this should be ready to merge, right?

PR #136, opened 2021-10-15
Update subproject for smaller system sizes

### LINKED ISSUE #135: Changing the water and ethanol box sizes
In our meeting, we discussed and agreed on the following changes to the project:
- Water box size change to ~32 Ang, around 1,000 waters.  
- Ethanol's molecules from 700 to 500.

I made the exact changes to the molecule count and box sizes in the Excel sheet, box_sizes tab.  In summary, they are:
- Water box side length = 32.07 Ang, with 1,100 water molecules
- Ethanol box side length = 36.46 Ang, with 500 ethanol molecules

Feel free to change these numbers slightly, but the MC folks, or at least I, like to keep a rounded number of molecules, as the MC steps are a multiple of the number of molecules.  I prefer rounded to 100.  For example, 1029 molecule yields weird output steps.

justinGilmer:
Closes #135

### DIFF
--- a/reproducibility_project/lrc_shift_subproject/subproject-init.py
+++ b/reproducibility_project/lrc_shift_subproject/subproject-init.py
@@ -40,77 +40,57 @@ for key in molecules:
 g_per_cm3 = u.g / (u.cm * u.cm * u.cm)
 masses = {
     "methaneUA": [16.04] * u.amu,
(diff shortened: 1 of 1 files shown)


bc118:
For the sub project, I thought we were only running water and not ethanol for the electro statics?  Or am I mistaken?  

Also we will need to change the standard/regular init.py file as well

PR #141, opened 2021-10-25
Analysis: works on lrcProject, 

dyukovsm:
aggregation flaw that doesnt let analysis run before all is finished still needs a fix - comming soon


dyukovsm:
Don't merge yet. I need to something.

PR #144, opened 2021-11-14
GOMC will now auto pass the set ensemble equilibrium if it runs all 3 cycles, regardless if it passes the pymbar test or not.

bc118:
GOMC will now auto pass the set ensemble equilibrium if it runs all 3 cycles, regardless if it passes the pymbar test or not.

This will auto-fix any stuck equilb runs in the sub-project and be used for the main project as well.


justinGilmer (approved):

justinGilmer:
Do you need this for the subproject too @bc118 ? Or just the main project?

bc118:
This is for the sub-project and the main project

PR #152, opened 2021-11-30
Mc3s final

ramanishsingh:
Updated code on lrc_subproject


jennyfothergill (commented):
I am not familiar with Mc3s, so I can't really comment on those files, but do you want to commit the results txt files?

ramanishsingh:
Yeah, there's not much need to commit the results. I need to git rm these files and then commit. right?

jennyfothergill:
> Yeah, there's not much need to commit the results. I need to git rm these files and then commit. right?

Yep! I think you can use `git rm --cached <filename>` to untrack it but not delete the file as well. You may also want to remove `
reproducibility_study/reproducibility_project/lrc_shift_subproject/src/engine_input/mcccs/methaneUA/NPT/.topmon.inp.swp `

ramanishsingh:
I am not able to find that file on my local copy of the repo!

jennyfothergill:
Yeah, it looks like they have been deleted 👍

reproducibility_project/lrc_shift_subproject/src/engines/mcccs/project.py
```diff
@@ -311,9 +345,24 @@ def sanitize_gemc_log(step):
     arrays_box1[:, 4] = arrays_box1[:, 4]  # Pressure kPa to kPa
     density_array = arrays_box1[:, 5] / (arrays_box1[:, 0]) ** 3
     density_array = density_array.reshape(density_array.shape[0], 1)
+
+    timestep = np.copy(density_array)
+    for i in range(timestep.shape[0]):
+        timestep[i] = i
```
jennyfothergill, line 351:
```suggestion
    timestep = np.arange(stop=density_array.shape[0])
```

jennyfothergill (approved):
A couple minor suggestions. When working with named arrays I might recommend [recfunctions](https://numpy.org/devdocs/user/basics.rec.html), but this is not necessary to hold up this PR.

PR #174, opened 2022-03-03
Modify pentaneUA statepoints for flexible and constrained

CalCraven:
In order to compare between MD and MC treatments of pentaneUA, we need to have both rigid and flexible models for pentaneUA.

	*Lammps-VU and lammps-UD can not do constrained bonds
	*All MC engines and HOOMD + Gromacs can do constrained bonds
	*All MC engines cannot do flexible bonds
	*All MD engines can do flexible bonds

This PR updates the statepoints from init.py with two pentane "molecules"; 1 that will be treated flexibly, and one that will be treated constrained.

### DIFF
--- a/reproducibility_project/init.py
+++ b/reproducibility_project/init.py
@@ -15,7 +15,14 @@ def dict_product(dd):
         yield dict(zip(keys, element))
 
 
-molecules = ["methaneUA", "pentaneUA-flexible_bonds", "pentaneUA-constrain_bonds", "benzeneUA", "waterSPCE", "ethanolAA"]
+molecules = [
+    "methaneUA",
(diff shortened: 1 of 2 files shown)


justinGilmer (approved):
I worked with @CalCraven on this PR, LGTM to me, but we need some additional folks to verify.

CalCraven:
I looked through the rest of the files, and I think we're good now. I added pentaneUA : Pentane() to the get_molecule function in system_builder, so it should work regardless of the name for the pentane molecule, such as the one used in the spe-subproject. Good catch @jennyfothergill

jennyfothergill (approved):
LGTM :)

daico007 (approved):

PR #175, opened 2022-03-03
Update benzeneUA ff and utils to TraPPE parameters 

justinGilmer:
Previously, we were using a benzene force field that we meant to be ran
as a flexible model, this led to inconsistent simulations between MC and
MD.

In addition, the flexible models had a very long equilibration time.

This switches the model to the standard TraPPE-UA representation of
benzene, which is treated as a rigid body.


bc118 (approved):
Looks like it works for me.  Note I am overriding the GOMC bond and angle K constants and not reporting the K constant energies.  Therefore, no bond, angle, or dihedral will be reported from GOMC.

As long as MD can handle the low bond and angle constants it should be OK.  GOMC is fine with any bobas and angle constants, as we override them to fix the bonds and angles.

CalCraven (approved):
LGTM. Just a comment, but should we consider in Foyer making it possible to set bond and angle constants to 0?

bc118:
I think this is questionable.  If a reasonable bond constant is not provided and an engine does not fix the molecule totally rigid,  I think you will run into a few issues.   I would leave it as default can’t be zero.

CalCraven:
To give a counter argument, I think just because it's questionable simulation practice and some of the engines don't support it, doesn't mean we should make it impossible to do for all engines. I'd argue that there's many just as questionable practices that we can allow, such as allowing bond constants that are extremely high that no engine can support. 

In my purview, it's not our job to be the arbitrators of what is possible, that's on the individual engines to yell at you if you did something that they can't handle. In this case, we want them to yell at you if these parameters are being used because they shouldn't be, which is a fringe case that it would be great for MoSDeF to be able to support.

bc118:
@CalCraven That is a fair argument.  I will add to your thought.  If the utilized bond constant of the molecule (bonds more, but also angles) not rigid, the system is likely to fail as nothing is holding it together.  Likely ending in an exploding system.  As long as the constants are not negative, which they never should be.

CalCraven:
Yeah and I can't think of a use case where negative values would want to be used. I'm also not certain exactly why setting things to 0 causes the failure. If it's due to some divide by 0 in the python code, that would be a good enough reason to disallow it because I don't want to be running around with a bunch of side cases checking for a few extremely uncommon possible use cases. But if we're manually checking for it to be 0 and throwing up the error on our side, that might be worth revisiting.

jennyfothergill (approved):
LGTM--in order to run this as a rigid body, hoomd will need to remove bond and angle forces. In response to @CalCraven's comment:

>LGTM. Just a comment, but should we consider in Foyer making it possible to set bond and angle constants to 0?

This could work well for a hoomd use-case. :) e.g. the bonds/angle forces are not returned in the forcefield.

bc118:
@jennyfothergill i don’t think they can be zero with the existing checks (unless removed),  but they may be able to be 1e-14 or something.  GOMC is manually overwriting the bonds and angle constants, so it should not matter to us.

daico007 (approved):

PR #189, opened 2022-05-04
GOMC analysis update

dyukovsm:
using code Brad pushed 2 days age updated my analysis and tested on the complete GOMC data set


bc118:
@dyukovsm This code needs throughly tested to make sure it is correct, working properly for the whole simulations from start to finish, and passing all tests before it is reviewed and merged.  This means rerunning simulation test cases and comparing .  Definitely rerunning methane and comparing to the current results with this exact code, and ensuring it matches.

reproducibility_project/src/engines/gomc/project.py
```diff
@@ -576,24 +578,25 @@ def part_3d_output_production_run_started(job):
...
-                    split_move_line = line.split()
-                    if (
-                        split_move_line[0] == "Move"
-                        and split_move_line[1] == "Type"
-                        and split_move_line[2] == "Mol."
-                        and split_move_line[3] == "Kind"
-                    ):
-                        job_run_properly_bool = True
-    else:
+    with job:
```
bc118, line 581: Why do we need this with job statement here?

bc118 (changes requested):

reproducibility_project/src/engines/gomc/project.py
dyukovsm: With statement is needed because when this test is used with aggregator, test doesn't see the file it's supposed to check.

dyukovsm:
I want to abandon this pull request and make a better one with all the requested changes made.

bc118:
You can just modify the code on this one.

PR #190, opened 2022-05-11
GOMC analysis update: rerun and tested mean energy, pressure and density values of NPT methane.

dyukovsm:


dyukovsm:
attached two files that verify that the added analysis code does not change the outcome of the methane run compared to when the methane run, when the analysis code was run after the simulation. Mean energy, pressure and density values agree to 6+ digits between each other and values provided by reproducibility planning excel.

[rclonedData.zip](https://github.com/mosdef-hub/reproducibility_study/files/8672590/rclonedData.zip)
[startToFinishData.zip](https://github.com/mosdef-hub/reproducibility_study/files/8672592/startToFinishData.zip)

PR #191, opened 2022-05-11
Updated GOMC analysis.

dyukovsm:
Previous merge request was automatically deleted because I couldn't find a quick way to not include a file in a pull request so I deleted and reforked the repository. 

https://github.com/mosdef-hub/reproducibility_study/pull/190

I attached the files that show that analysis does not impact the simulation by running simulation without analysis (rcloned) and with analysis (startToFinish)

[startToFinishData.zip](https://github.com/mosdef-hub/reproducibility_study/files/8673063/startToFinishData.zip)
[rclonedData.zip](https://github.com/mosdef-hub/reproducibility_study/files/8673064/rclonedData.zip)
)


dyukovsm:
Updated data from 16 replicates of methane run:

[setAverages.txt](https://github.com/mosdef-hub/reproducibility_study/files/8758592/setAverages.txt)

bc118:
I will approve this PR.  The only little thing I noticed is that when adding replicates, the analysis will not run for the new replicates unless the original replicate's analysis files are deleted.  I don't think this should be an issue for anyone, but thought it was worth noting in the comments.

bc118 (approved):
I will approve this PR. The only little thing I noticed is that when adding replicates, the analysis will not run for the new replicates unless the original replicate's analysis files are deleted. I don't think this should be an issue for anyone, but I thought it was worth noting in the comments.

CalCraven (approved):
Thanks for writing these checks @dyukovsm

PR #194, opened 2022-09-02
Mc3s final

ramanishsingh:
Update pentane connstrain bond sp


CalCraven:
Not sure why this PR is causing the test to fail, something with the RDF analysis. Since we're just using a predetermined number to compare to, we need to look into what the correct answer is to validate if the test is wrong or something in this PR is causing that value to have changed.

CalCraven (commented):
Pinning some of these versions should also fix the failing test using freud, that I'm guessing is due some changes in that package as well. I have freud 2.6.2, but the most recent is freud 2.11.0

daico007:
Checking this PR out now, definitely related to version of freud, working on updating test to reflect the changes.

PR #198, opened 2022-09-11
add mdmc_flexbond files and analysis files

ramanishsingh:


CalCraven (changes requested):
Some minor changes to format to match the main project. I'd recommend also making an aggregate summary subproject level directory to put the summary density plots that we will compare to the main project densities. I'm not 100% sure how we will generate this data unless we use a dropbox method to aggregate it, or our simulation data is small enough to rsync over to a local cluster and run that script.

CalCraven (approved):
After further discussion, we're going to leave the individual project as is, with single engine data outputs compiled on local servers and uploaded in the /src directory, and then compare across engines afterwards.

PR #209, opened 2023-01-27
Lammps NPT-fixOH jobs analysis

CalCraven:
This PR contains the Job files used for the LAMMPS jobs used for fixed bond lengths. It includes the analysis for:
1. Densities
2. Bond lengths with the molecules
3. RDF
4. Hydrogen bonding


daico007:
Seems like there might be  some issue with the hbond calculation (everything is 0)

PR #216, opened 2023-02-17
Add spce water with nist params subproject

daico007:
Add new subproject concerning water SPCE with updated params from NIST for gromacs and lammps. @CalCraven @ramanishsingh Can y'all review the number and see if it's correct? Feel free to push to this PR any change you see fit.


ramanishsingh:
The parameters in the waterSPCE_nist.xml file look correct.

reproducibility_project/src/xmls/waterSPCE_nist.xml
```diff
@@ -0,0 +1,22 @@
...
+   <Bond class1="OW" class2="HW" length="0.100" k="345000.0"/>
+ </HarmonicBondForce>
+ <HarmonicAngleForce>
+   <!-- https://github.com/gromacs/gromacs/blob/master/share/top/oplsaa.ff/spce.itp -->
+   <!-- numpy.deg2rad(109.47).round(8) -->
+   <!-- numpy.__version__ 1.21.1 -->
+   <Angle class1="HW" class2="OW" class3="HW" angle="1.91061193" k="383.0"/>
+ </HarmonicAngleForce>
+ <NonbondedForce coulomb14scale="0.5" lj14scale="0.5">
+   <!-- https://github.com/gromacs/gromacs/blob/master/share/top/oplsaa.ff/spce.itp -->
```
CalCraven, line 18: This source is wrong, should source the spce website.

reproducibility_project/src/xmls/waterSPCE_nist.xml
```diff
@@ -0,0 +1,22 @@
...
+  <Atom type="opls-116" charge="-0.8476" sigma="0.316555789" epsilon="0.650165504466"/>
```
CalCraven, line 19: This number looks wrong to me:
I converted by hand:
`K = 78.19743111`
`K * 1.380649e-23 / 1000 * 6.02214076e23`
`0.6501696177997083`

And converted using scipy / unyt:
```python
conversion_quantity = u.unyt_quantity(1, "K").to_value(
    "kJ/mol", equivalence="thermal"
)
K = 78.19743111
K * conversion_quantity
```
`0.6501696177997083`

In summary, I'm not certain I trust https://www.colby.edu/chemistry/PChem/Hartree.html for this high precision conversion.

CalCraven (changes requested):
If anyone can double check my conversions or show me something that was done wrong, let me know. Really want to make sure we get this right this time around.

reproducibility_project/waterspce_nist_subproject/init.py
```diff
@@ -0,0 +1,169 @@
...
+masses = {
+    "waterSPCE": [18.015324] * u.amu,
+}
+init_density_liq = {
+    "waterSPCE": [0.998] * g_per_cm3,
+}
+init_density_vap = {
+    "waterSPCE": [None],
+}
+temperatures = {
```
CalCraven, line 55: Do we want to run this many simulations? We can probably get away with just running at 300K I think, especially if the results look good.

reproducibility_project/src/xmls/waterSPCE_nist.xml
daico007: Good spot, sorry, I was copy the other one over and forgot to change the source

reproducibility_project/waterspce_nist_subproject/init.py
daico007: I was thinking running the whole set can help you see if the small difference could create different effect at different temperature, but I am down to running just one

reproducibility_project/src/xmls/waterSPCE_nist.xml
daico007: I am torn on this one, @ramanishsingh what do you think?

daico007: @chrisiacovella

reproducibility_project/src/xmls/waterSPCE_nist.xml
```diff
@@ -0,0 +1,22 @@
+<ForceField version="0.0.1" name="SPC/E Water" combining_rule="geometric">
+ <AtomTypes>
+   <Type name="opls-116" class="OW" element="O" mass="15.99940" def="[O;X2](H)(H)" desc="SPC/E Oxygen" doi="10.1021/j100308a038"/>
+   <!-- Need to validate the Hydrogen SMARTS string below -->
```
daico007, line 4:
```suggestion
```

reproducibility_project/src/xmls/waterSPCE_nist.xml
daico007, line 18:
```suggestion
   <!-- https://doi.org/10.1021/j100308a038 -->
```

reproducibility_project/waterspce_nist_subproject/init.py
CalCraven: For LAMMPS, presumably the slower engine, the production run time was Total wall time: 10:41:30 for 5 ns. It took about 7 ns to equilibrate so that would put us at about 24 hours per simulation, so we could presumably finish these in about 2-3 days for the 16 replicates, assuming low queue times.

reproducibility_project/src/xmls/waterSPCE_nist.xml
CalCraven: Boltzmann constant source: https://www.nist.gov/si-redefinition/kelvin-boltzmann-constant
Avogdros Number source: https://physics.nist.gov/cgi-bin/cuu/Value?na

reproducibility_project/waterspce_nist_subproject/init.py
daico007: the queue is pretty empty (for now). I have been running the gromacs stuff, though with the slightly diff params again. I can stop those and rerun when we agree upon a final number

reproducibility_project/src/xmls/waterSPCE_nist.xml
CalCraven: From the Berendsen Paper:
A = 0.37122 ((kJ/mol)^1/6)*nm and B = 0.3428 (kJ/mol)^1/12*nm
A^12=4ϵσ12,  B^6=4ϵσ6 xx

sigma = 0.31655578901998815 nm 
epsilon = 0.6501695808187479 kJ/mol


I would suggest these are the values we use, if we want to be exact. However, if we want to use the NIST values, then we should use the ones I've hand converted. Both the unyt conversion from the exact NIST thermal units and the hand calculation gives the same value, as opposed to the colby one which we have no exact idea how that conversion is done.

ramanishsingh: @CalCraven  These numbers make sense. I am also getting the same.
I got these numbers: 
`
sigma is 3.165557890199881401361442585 Angstrom
epsilon is 0.6501695808187480926657690985 kJ/mol = 0.1553942592779034577606330816 kcal/mol = 78.19742666221283017109788421 K`

I used the Decimal package in Python to do the math. 

`#U = -(A/r)^6+(B/r)^12
#that makes
#sigma = B^2/A and epsilon = (1/4)*(A/B)^12

from decimal import *
A = Decimal(0.37122)# ((kJ/mol)^1/6)nm
B = Decimal(0.3428) #(kJ/mol)^1/12nm
kB = Decimal(1.380649 *1e-23)# J/K
NA = Decimal(6.02214076*1e23) #1/mol

sigma = Decimal(10)*B**2/A #Ang
epsilon_kJmol = Decimal(0.25)*(A/B)**12 #kJ/mol

epsilon_kcalmol = epsilon_kJmol/Decimal(4.184)

epsilon_K = Decimal(1000)*epsilon_kJmol/(NA*kB)


print("sigma is {} Angstrom".format(sigma))
print("epsilon is {} kJ/mol = {} kcal/mol = {} K".format(epsilon_kJmol, epsilon_kcalmol, epsilon_K))
`

PR #217, opened 2023-02-21
remove null ensemble

daico007:
There was a mistake in the subproject `init.py` that create double the necessary job.


CalCraven (approved):
Thanks for the speedy change!

PR #218, opened 2023-02-21
Adjustments to project.py for LAMMPS Simulations

CalCraven:
* Copy over rahman project.py for running on slurm rahman
* Serial signac submission template adjustments
* Add in a new lammps spce forcefield xml to run for gromacs and lammps at 300K
* Number of jobs in water_spce subproject = 128


daico007 (approved):
LGTM!

PR #220, opened 2023-03-02
Cassandra spe water rerun

rwsmith7531:
Cassandra had different results for waterSPCE the SPE subproject compared to other simulation engines due to rounding the LJ parameters to only 3 decimal places (4 significant digits for sigma) despite more significant digits (6 for sigma) being present in the source FF xml file and used by other simulation engines.  This PR includes the results from rerunning the SPE subproject for waterSPCE after locally increasing the number of decimal places to 5 for the LJ parameters written by mbuild's MCF writer.  This yields SPE results that are very consistent with other simulation engines.


emarinri:
Looks good! Thanks!

CalCraven (approved):
LGTM!

daico007 (approved):

PR #234, opened 2023-04-10
Repair gomc log-spe.txt

CalCraven:
Functions and data saved out for all job log-spe.txt in single point energy subproject.


CalCraven:
I think this PR is going to be redone in a separate PR that reads directly from the GOMC files.

PR #246, opened 2023-06-05
Set up new simulation for ethanol 

daico007:
Per last week discussion with @CalCraven, run extra ethanol simulation with HOOMD4 using the Bussi thermostat.


daico007:
Replaced by #250

PR #250, opened 2023-08-09
HOOMD-blue 4.x subproject

joaander:
Updates to the hoomd4 subproject.

The hoomd4 simulations were run on NCSA Delta which requires the latest version of signac-flow which in turn requires signac>= 2.0. I needed to make some changes outside the subproject to account for the missing `signac.contrib.job` in signac 2.0. Only the hoomd4 project file has been updated for the latest flow, so it gets a subproject specific environment file.

I also removed the flow job template scripts. These are highly environment specific and should be created by the end user as needed.


joaander:
Thanks to @daico007 for much assistance putting this together.

daico007 (approved):

PR #266, opened 2024-09-21
update MWs in MCCCS-MN log file generation

ramanishsingh:


CalCraven (approved):
Should be good to merge this.

